PR #2, opened 2020-05-22
Circleci project setup

eoln:
setup is derived from the central-ledger project,
build and deploy jobs are temporarily commented out


Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apk add --no-cache -t build-dependencies git make gcc g++ python libtool autoconf automake \
+    && cd $(npm root -g)/npm \
+    && npm config set unsafe-perm true \
+    && npm install -g node-gyp
+
+COPY package.json package-lock.json* /opt/auth-service/
+
+RUN npm ci
+COPY src /opt/auth-service/src
+COPY test /opt/auth-service/
```
lewisdaly, line 13: Do we need this here?

eoln: we need `src` to be able to run the typescript source code with ts-node
we need `test` to be able to run the tests inside the container

lewisdaly: Sure. We actually don't normally run the unit tests inside of our containers, but outside before we build them. That way we don't need the `devDependencies` as well.

I'm happy to leave this for now, and we can discuss later on.

PR #6, opened 2020-06-23
Add Entity Relationship Diagram and Attribute Table to Documentation

spikerheado1234:
Addresses the first task for #341.

Consists of the entity relationship diagram describing the database schema. As well as a table of attributes describing internal representation of their types within the database alongside any constraints they have.


docs/README.md
```diff
@@ -1,6 +1,24 @@
...
+
+<p align="center">
+  <img src="ErDiagram.png">
+</p>
+
+The following table depicts the types of each attribute in our table, algonside their constraints:
+
+|Attribute| Represented Internal Type | Constraints|
+|:---------:|:---------------------------:|:------------:|
+|`ID`|BIGINT|Unique - Primary Key|
```
lewisdaly, line 16: A couple notes to bring this into line with other Mojaloop standards:
- `ID` should be `id`, and of type `VARCHAR(36)` 
  - take a look at [central-ledger](https://github.com/mojaloop/central-ledger/blob/master/migrations/300200_transfer.js#L32) for example (but don't copy the `tableNameId` pattern, `id` is much better)
  - Also, just for reference: https://github.com/mojaloop/mojaloop-specification/blob/master/Supporting%20Files/fspiop-rest-v1.0-OpenAPI-implementation.yaml#L3061

- Tablenames in the database should be snake case: e.g. `initiator_id`, `participant_id`
  - Knex can map from `snake_case` to `camelCase` for us
- `initiator_id` and `participant_id` should probably also be `VARCHAR(32)`, according to the [swagger spec](https://github.com/mojaloop/mojaloop-specification/blob/master/Supporting%20Files/fspiop-rest-v1.0-OpenAPI-implementation.yaml#L3282)

- Can you go into more detail (perhaps in a different table) about the format of `scopes` and `credential`? I thought that `credential` would be expanded into separate, _nullable_ columns based on our conversation the other day, eg:
```
credential_id:
credential_status: 
credential_type:
credential_challenge:
credential_payload:
```
Alternatively, this could also just be a _nullable_ JSON field as well.

- Finally (and I'm really nitpicking here) the table should probably be called `consent`, and not `Consents`

spikerheado1234: Addressed, and modified, the PR.

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+|Attribute| Represented Internal Type | Constraints|
+|:---------:|:---------------------------:|:------------:|
+|`id`|VARCHAR(36)|Unique - Primary Key|
+|`initiator_id`|VARCHAR(36)|Non-Null|
+|`participant_id`|VARCHAR(36)|Non-Null|
+|`credential_id`|TEXT|(Can be Null)|
+|`credential_type`|TEXT|(Can be Null)|
+|`credential_status`|TEXT|(Can be Null)|
+|`credential_payload`|TEXT|(Can be Null)|
+|`credential_challenge`|TEXT|(Can be Null)|
```
jgeewax, line 23:
```suggestion
|`initiator_id`|`VARCHAR(36)`|Non-Null|
|`participant_id`|`VARCHAR(36)`|Non-Null|
|`credential_id`|`TEXT`|(Can be Null)|
|`credential_type`|`TEXT`|(Can be Null)|
|`credential_status`|`TEXT`|(Can be Null)|
|`credential_payload`|`TEXT`|(Can be Null)|
|`credential_challenge`|`TEXT`|(Can be Null)|
```

jgeewax: Just adding code ticks around these.

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+</p>
+
+The following table depicts the types of each attribute in the `Consent` table, algonside their constraints:
+
+|Attribute| Represented Internal Type | Constraints|
+|:---------:|:---------------------------:|:------------:|
+|`id`|VARCHAR(36)|Unique - Primary Key|
+|`initiator_id`|VARCHAR(36)|Non-Null|
+|`participant_id`|VARCHAR(36)|Non-Null|
+|`credential_id`|TEXT|(Can be Null)|
```
jgeewax, line 19: Can we double check on the format of this? I suspect that this would be numeric, but we should probably limit this rather than leave it as an unbounded text field

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+|`credential_type`|TEXT|(Can be Null)|
```
jgeewax, line 20: This is determined by us and should be limited rather than an arbitrary text field. Maybe `VARCHAR(32)` ?

docs/README.md
jgeewax, line 23: `credential_challenge` is something we can determine the size of. Why not decide on how many bytes we care about and then just go with that? So `CHAR(128)` or something like that?

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+|`credential_status`|TEXT|(Can be Null)|
```
jgeewax, line 21: The status should be a set of specific values, so definitely not unbounded text. Maybe `VARCHAR(16)`?

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+|`credential_payload`|TEXT|(Can be Null)|
+|`credential_challenge`|TEXT|(Can be Null)|
+
+The following table depicts the types of each attribute in the `Scope` table:
+
+|Attribute| Represented Internal Type | Constraints|
+|:---------:|:---------------------------:|:------------:|
+|`id`|VARCHAR(36)|Unique - Primary Key|
+|`consent_id`|VARCHAR(36)|Foreign Key referring to the `id` in the consent table|
+|`scope_account_type`|VARCHAR(36)|Non-Null|
```
jgeewax, line 31:
```suggestion
|`action`|`VARCHAR(36)`|Non-Null|
```

jgeewax: This isn't really an account type, it's an action permitted on an account (e.g., `get balance` or `send money`)

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+|`credential_status`|TEXT|(Can be Null)|
+|`credential_payload`|TEXT|(Can be Null)|
+|`credential_challenge`|TEXT|(Can be Null)|
+
+The following table depicts the types of each attribute in the `Scope` table:
+
+|Attribute| Represented Internal Type | Constraints|
+|:---------:|:---------------------------:|:------------:|
+|`id`|VARCHAR(36)|Unique - Primary Key|
+|`consent_id`|VARCHAR(36)|Foreign Key referring to the `id` in the consent table|
```
jgeewax, line 30:
```suggestion
|`consent_id`|`VARCHAR(36)`|Foreign Key referring to the `id` in the consent table|
```

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+|`scope_account_type`|VARCHAR(36)|Non-Null|
+|`scope_account_number`|VARCHAR(36)|(Can be Null)|
```
jgeewax, line 32:
```suggestion
|`account_number`|`VARCHAR(36)`|(Can be Null)|
```

docs/README.md
```diff
@@ -1,6 +1,38 @@
...
+
+The reason attributes pertaining to `credentials` and `scope_account_number` can be null is because of the communication sequence established in between the `PISP` and the `authorisation-service`. Where, at the point of first contact with the authorisation-service, these fields are not present. They are available during later parts in the sequence of communication in between the two services.
```
jgeewax, line 34:
```suggestion
The reason attributes pertaining to `credentials` and `account_id` can be null is because of the communication sequence established in between the PISP and the auth service. At the point of first contact with the auth service, these fields are not present and instead are added later on in the sequence of communication in between the two services.
```

docs/README.md
ramanmangla: Modified data types and updated the PR.

docs/README.md
ramanmangla: The status is either `PENDING` or `ACTIVE`. So `VARCHAR(10)` should be good? 
Updated the PR.

docs/README.md
ramanmangla: Formatted the data types and updated the PR.

docs/README.md
ramanmangla: Isn't the credential info populated by the Auth service itself? If my understanding of the sequence diagrams is correct, the service itself creates and populates all the credential attributes, except the public key (payload), when it receives the request from the switch. Thus, the only credential attribute that should be allowed to be NULL is the payload.

The scopes and the account numbers are received by the Auth service in the incoming request too. If there are no scopes, we don't need to add any row in the Scopes table. Hence, they can be Non-Nullable too.

spikerheado1234: Yes, I agree. The only nullable value should theoretically be `challenge_payload`.

spikerheado1234:
> ER diagrams need to be updated to reflect the schema changes.

Updated. For the future, you can also manually update them. The file lies in docs/ErDiagram.puml. You can change the fields here, and it will automatically re-render the correct image in the repo.

lewisdaly:
@spikerheado1234 when this is ready for review, you can can click the 're-request review' button in the top right.

docs/README.md
```diff
@@ -1,6 +1,40 @@
...
+|`credential_challenge`|`CHAR(128)`|Non-Null|Base 64 encoded challenge string|
+
+The following table depicts the types of each attribute in the `Scope` table:
+
+|Attribute| Represented Internal Type | Constraints| Description |
+|:---------:|:---------------------------:|:------------:|:---------------------------:|
+|`id`|`VARCHAR(36)`|Unique - Primary Key|Alphanumeric Scope ID|
+|`consent_id`|`VARCHAR(36)`|Non-Null - Foreign Key|Foreign Key for `id` in the `Consent` table|
+|`action`|`VARCHAR(36)`|Non-Null|Account scope (enum) action allowed with corresponding consent|
+|`account_number`|`VARCHAR(36)`|Non-Null|Account number for associated scope|
```
lewisdaly, line 32: I think this should be `account_id`, and not account number. The `account_id` is an opaque identifier provided by the DFSP, and calling it 'number' is a little too opinionated, in my opinion ;)

docs/README.md
```diff
@@ -1,6 +1,40 @@
...
+</p>
+
+The following table depicts the types of each attribute in the `Consent` table along with the constraints:
+
+|Attribute| Represented Internal Type | Constraints| Description |
+|:---------:|:---------------------------:|:------------:|:---------------------------:|
+|`id`|`VARCHAR(36)`|Unique - Primary Key|Alphanumeric consent ID|
+|`initiator_id`|`VARCHAR(32)`|Non-Null|Alphanumeric PISP ID|
+|`participant_id`|`VARCHAR(32)`|Non-Null|Alphanumeric DFSP ID|
+|`credential_id`|`INT UNSIGNED`|Unique - Non-Null|Unsigned numeric ID - upto 2^32 records|
```
lewisdaly, line 19: I think this **needs** to be _nullable_, since it will be filled in by the PISP - it's the unique identifier of the FIDO key on the device, so won't be filled in until the PISP has signed the challenge.

docs/README.md
```diff
@@ -1,6 +1,40 @@
...
+|`credential_type`|`VARCHAR(16)`|Non-Null|Alphanumeric enum value|
+|`credential_status`|`VARCHAR(10)`|Non-Null|String - `PENDING`/`ACTIVE`|
+|`credential_payload`|`TEXT`|Can be Null|Public key string|
+|`credential_challenge`|`CHAR(128)`|Non-Null|Base 64 encoded challenge string|
```
lewisdaly, line 23: A _Non-Null_ constraint here implies that the `credential_challenge` is created when the `Consent` object is created. 

However, in [1.5-grant-consent](https://github.com/jgeewax/pisp/blob/master/docs/linking/README.md#15-grant-consent), the credential object is still null, and the challenge isn't created until `CRED-5` in [1.6.1 Requesting a challenge](https://github.com/jgeewax/pisp/blob/master/docs/linking/README.md#161-requesting-a-challenge)

docs/README.md
ramanmangla: Is there a specific data type for the PISP credential ID or can the auth service specify the type of ID it wants?

docs/ErDiagram.puml
```diff
@@ -0,0 +1,30 @@
+@startuml
+
+title Entity Relationship Diagram 
+' hide the spot
+hide circle
+
+' avoid problems with angled crows feet
+skinparam linetype ortho
+
+entity "Consent" as e01 {
```
jgeewax, line 10: I think this can just be `entity Consent {` and then `Consent ||..o{ Scope` down below. Might be easier to read.

docs/ErDiagram.puml
```diff
@@ -0,0 +1,30 @@
...
+  * **id**
+  * initiator_id
```
jgeewax, line 12: Can we add a field for the time the consent resource was created? Just for history...

docs/ErDiagram.puml
```diff
@@ -0,0 +1,31 @@
+@startuml
+
+title Entity Relationship Diagram 
+' hide the spot
+hide circle
+
+' avoid problems with angled crows feet
+skinparam linetype ortho
+
+entity "Consent" {
```
jgeewax, line 10:
```suggestion
entity Consent {
```

docs/ErDiagram.puml
```diff
@@ -0,0 +1,31 @@
...
+  * participant_id
+  * time_stamp
+  * credential_id
+  * credential_type
+  * credential_status
+  * credential_payload
+  * credential_challenge
+}
+
+entity "Scope" {
```
jgeewax, line 22:
```suggestion
entity Scope {
```

docs/README.md
```diff
@@ -1,6 +1,40 @@
...
+</p>
+
+The following table depicts the types of each attribute in the `Consent` table along with the constraints:
+
+|Attribute| Represented Internal Type | Constraints| Description |
+|:---------:|:---------------------------:|:------------:|:---------------------------:|
+|`id`|`VARCHAR(36)`|Unique - Primary Key|Alphanumeric consent ID|
+|`initiator_id`|`VARCHAR(32)`|Non-Null|Alphanumeric PISP ID|
+|`participant_id`|`VARCHAR(32)`|Non-Null|Alphanumeric DFSP ID|
+|`credential_id`|`INT UNSIGNED`|Nullable|Unsigned numeric ID - upto 2^32 records|
```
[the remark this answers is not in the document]
jgeewax: https://www.w3.org/TR/credential-management-1/#credential says webauthn uses a string

android apis seem to use a string also https://developers.google.com/android/reference/com/google/android/gms/fido/fido2/api/common/PublicKeyCredential#getId()

docs/README.md
```diff
@@ -1,6 +1,43 @@
...
+
+The following table depicts the types of each attribute in the `Consent` table along with the constraints:
+
+|Attribute| Represented Internal Type | Constraints| Description |
+|:---------:|:---------------------------:|:------------:|:---------------------------:|
+|`id`|`VARCHAR(36)`|Unique - Primary Key|Alphanumeric consent ID|
+|`initiator_id`|`VARCHAR(32)`|Non-Null|Alphanumeric PISP ID|
+|`participant_id`|`VARCHAR(32)`|Non-Null|Alphanumeric DFSP ID|
+|`created_at`|`TIMESTAMP`|Non-Null|Timestamp of consent creation|
+|`credential_id`|`CHAR(256)`|Nullable|Unsigned numeric ID - upto 2^32 records|
```
jgeewax, line 20: https://codelabs.developers.google.com/codelabs/fido2-for-android/#5 also shows a string being used.

```suggestion
|`credential_id`|`VARCHAR(256)`|Nullable|Unsigned numeric ID - upto 2^32 records|
```

PR #8, opened 2020-06-25
feat(swagger): Add new Auth Service endpoints to swagger.json

lewisdaly:
- Also splits swagger references into separate yaml files
> Note: I tried to update to Swagger/OpenAPI 3.0, and use the `api-snippets` library, but this caused a whole lot of other issues as `hapi-openapi` doesn't support the v3.0 specification

Part of #280


package.json
```diff
@@ -18,6 +18,7 @@
     "release": "standard-version",
     "standard": "echo '\\033[1;33m  This project uses eslint instead of standard. Use `npm run lint` instead.'",
     "start": "ts-node -P ./tsconfig.json ./src/cli.ts",
+    "dev": "nodemon --watch 'src/**/*.ts' --exec 'ts-node' src/cli.ts",
```
eoln, line 21: if we already used ts-node, why nodemon instead of ts-node-dev? https://www.npmjs.com/package/ts-node-dev

lewisdaly: Ah I didn't know it existed!

lewisdaly: Good point, I didn't know it existed

lewisdaly: @eoln I might leave this for now - I tried switching it out, but `ts-node-dev` has an annoying "feature" where it rings the bell and sends a notification on every change... and this "feature" can't be switched off

PR #9, opened 2020-06-30
[WIP] #341 Set up migrations and seeds and SQLite Testing Environment

spikerheado1234:
In reference to ticket number #341 . Addresses the second and fourth tasks for it. 

Consists of setting up the migrations and sample seeds for the local dockerized mysql instance as well as setting up an in-memory SQLite for unit and integration testing. I have put in some Unit tests as well for the seeds, verifying that the basic constraints on the table and seed files are executed correctly.


spikerheado1234:
Please ignore the files in the `/docs/` folder, the other PR is not merged into master yet and thus the diff is showed within the PR (Poor git practices on my part). 

Just wanted to open up this PR because it is large and will likely require multiple revisions of feedback.

migrations/20200624121732_consent.ts
```diff
@@ -0,0 +1,47 @@
...
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Ahan Gupta <ahangupta.96@gmail.com>
+
+ --------------
+ ******/
+import * as Knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void> => {
+  return await knex.schema.hasTable('Consent').then(function (exists: boolean): Knex.SchemaBuilder | void {
```
lewisdaly, line 28: Minor nitpicks 
1. No need to `await` if you are returning the promise
2. `.then` should be on a new line, and
3.  please use `() => {}` syntax where possible



```suggestion
  return knex.schema.hasTable('Consent')
    .then((exists: boolean): Knex.SchemaBuilder | void  => {...})
```

migrations/20200624121732_consent.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (!exists) {
```
lewisdaly, line 29: A general comment about style: it tends to be easier to read if you just exit early and make the _main thing you want to do_ be at the end of the function, e.g.

```suggestion
   //e.g. take care of negative cases first
    if (exists) {
      return;
    }

    // This leaves room for other conditional handling in the future here
    
    //Then  we can do the thing that we're all here to do
    return knex.schema.createTable(...)

```

migrations/20200624121732_consent.ts
```diff
@@ -0,0 +1,47 @@
...
+      return knex.schema.createTable('Consent', (t: Knex.CreateTableBuilder): void => {
+        t.string('id', 32).primary().notNullable()
+        t.string('initiatorId', 32).notNullable()
+        t.string('participantId', 32).notNullable()
+        t.timestamp('timeStamp').defaultTo(knex.fn.now())
```
lewisdaly, line 34: maybe this should just be `timestamp`? And perhaps this should be `createdAt`, since timestamp is a little ambiguous (maybe that's a comment for the other PR...)

migrations/20200624121736_scope.ts
```diff
@@ -0,0 +1,44 @@
...
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Ahan Gupta <ahangupta.96@gmail.com>
+
+ --------------
+ ******/
+import * as Knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void> => {
+  return await knex.schema.hasTable('Scope').then(function (exists: boolean): Knex.SchemaBuilder | void {
```
lewisdaly, line 28: Same comments as above

seeds/01_consent.ts
```diff
@@ -0,0 +1,72 @@
...
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | number {
+  try {
+    return knex('Consent').insert(consents)
+  } catch (err) {
+    if (err.code === 'ER_DUP_ENTRY') return -1001
+    else {
+      console.log(`Uploading seeds for consents has failed with the following error: ${err}`)
+      return -1000
```
lewisdaly, line 69: This should probably throw a `new Error()` error instead of returning a number.

seeds/02_scope.ts
```diff
@@ -0,0 +1,60 @@
...
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | number {
+  try {
+    return knex('Scope').insert(scopes)
+  } catch (err) {
+    if (err.code === 'ER_DUP_ENTRY') return -1001
+    else {
+      console.log(`Uploading seeds for scopes has failed with the following error: ${err}`)
+      return -1000
```
lewisdaly, line 57: Same as above

test/unit/seeds/consent.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  it('should properly select all the entries in the Consent table', async (): Promise<void> => {
+    expect(db).toBeDefined()
+    const users: Knex.QueryBuilder[] = await db.from('Consent').select('*')
+    expect(users.length).toEqual(3)
+    expect(users[0]).toMatchObject({ id: '123', initiatorId: 'PISPA', participantId: 'DFSPA', credentialId: null, credentialType: null, credentialStatus: null, credentialPayload: null, credentialChallenge: null })
+    expect(users[1]).toMatchObject({ id: '124', initiatorId: 'PISPB', participantId: 'DFSPA', credentialId: 9876, credentialType: 'FIDO', credentialStatus: 'PENDING', credentialPayload: null, credentialChallenge: 'string_representing_challenge_a' })
+    expect(users[2]).toMatchObject({ id: '125', initiatorId: 'PISPC', participantId: 'DFSPA', credentialId: 9875, credentialType: 'FIDO', credentialStatus: 'VERIFIED', credentialPayload: 'string_representing_public_key_a', credentialChallenge: 'string_representing_challenge_b' })
+  })
+
+  it('should properly enforce the constraints in the Consent table', async (): Promise<void> => {
```
lewisdaly, line 52: These are great tests!

Minor note: we try and make  each `it` statement only test _1_ thing. So perhaps you can convert the `it('should properly enforce the constraints in the Consent table'` into a describe block, and then have a bunch of its:
```
it('enforces Primary Key Constraint')
it('enforces Non-Nullable constraint for initiatorId')
it('enforces Non-Nullable constraint for participantID')
...
```

test/unit/seeds/scope.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  it('should properly select all the entries in the Scope table', async (): Promise<void> => {
+    expect(db).toBeDefined()
+    const users: Knex.QueryBuilder[] = await db.from('Scope').select('*')
+    expect(users.length).toEqual(3)
+    expect(users[0]).toEqual({ id: 1, consentId: '123', action: 'accounts.getBalance', accountId: '12345-67890' })
+    expect(users[1]).toEqual({ id: 2, consentId: '123', action: 'accounts.transfer', accountId: '12345-67890' })
+    expect(users[2]).toEqual({ id: 3, consentId: '124', action: 'accounts.transfer', accountId: '21345-67890' })
+  })
+  it('should properly enforce all the constraints in the Scope table', async (): Promise<void> => {
```
lewisdaly, line 51: Same as above for multiple tests

migrations/20200624121732_consent.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Ahan Gupta <ahangupta.96@gmail.com>
+
+ --------------
+ ******/
+import * as Knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void> => {
```
eoln, line 27: shouldn't the return type be `Promise<void | Knex.SchemaBuilder>`? The returned value from line 28 is Knex.SchemaBuilder. The values returned by `then` callbacks are transformed into promises to allow chaining: https://javascript.info/promise-chaining

migrations/20200624121732_consent.ts
eoln: In Knex example given here: http://knexjs.org/#Schema-createTable is used `if(!exist)` version, but then for the `else` case the returned value is `undefined` in an implicit way. Explicit way always speeds up reading the code. This pattern also allows us to keep if/else nested blocks as flat as possible for complex cases. Let use the proposed by @lewisdaly implicit way.

migrations/migration.template
```diff
@@ -0,0 +1,9 @@
+import * as knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void> => {
+
+}
+
+exports.down = function(knex: Knex): Knex.SchemaBuilder {
```
eoln, line 7: shouldn't be a Promise for returned type for both `up` and `down`?

migrations/20200624121732_consent.ts
eoln: maybe `createTableIfNotExist` could be used: https://github.com/knex/knex/blob/master/types/index.d.ts#L1510

src/lib/config.ts
```diff
@@ -0,0 +1,68 @@
...
+ their names indented and be marked with a '-'. Email address can be added
+ optionally within square brackets <email>.
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Ahan Gupta <ahangupta.96@gmail.com>
+ --------------
+ ******/
+
+import rc from 'rc'
```
eoln, line 25: I've seen in last demo example of using Mozilla/node-convict module instead of rc: maybe we can migrate from rc? https://github.com/mozilla/node-convict

config/knexfile.ts
```diff
@@ -0,0 +1,40 @@
...
+      directory: seedsDirectory,
+      loadExtensions: ['.ts']
+    }
+  },
+  test: {
+    client: 'sqlite3',
+    connection: ':memory:',
+    useNullAsDefault: true,
+    migrations: {
+      directory: path.join(__dirname, '../migrations/'),
```
lewisdaly, line 28: should this just be `migrationsDirectory`?

config/knexfile.ts
```diff
@@ -0,0 +1,40 @@
...
+      tableName: 'auth-service',
+      loadExtensions: ['.ts']
+    },
+    seeds: {
+      directory: path.join(__dirname, '../seeds/'),
```
lewisdaly, line 33: This should be `seedsDirectory`

migrations/migration.template
lewisdaly, line 7: this should be same syntax as above, and maybe es6 exports as well

seeds/01_consent.ts
```diff
@@ -0,0 +1,72 @@
...
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | Error {
+  try {
+    return knex('Consent').insert(consents)
+  } catch (err) {
+    if (err.code === 'ER_DUP_ENTRY') return new Error('-1001')
+    else {
+      console.log(`Uploading seeds for consents has failed with the following error: ${err}`)
+      return new Error('-1000')
```
lewisdaly, line 69: Maybe this would be better as:
```suggestion
      throw err
```

lewisdaly: Unless you don't want to throw the error for some reason...

seeds/02_scope.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | Error {
+  try {
+    return knex('Scope').insert(scopes)
+  } catch (err) {
+    if (err.code === 'ER_DUP_ENTRY') return new Error('-1001')
+    else {
+      console.log(`Uploading seeds for scopes has failed with the following error: ${err}`)
```
lewisdaly, line 53: same as above, `throw err` might be more suitable

migrations/20200624121736_scope.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Ahan Gupta <ahangupta.96@gmail.com>
+
+ --------------
+ ******/
+import * as Knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void | Knex.SchemaBuilder> => {
```
eoln, line 27: same as for ..._consent.ts file - use ES Module syntax

seeds/01_consent.ts
```diff
@@ -0,0 +1,72 @@
...
+    participantId: 'DFSPA',
+    credentialId: '9875',
+    credentialType: 'FIDO',
+    credentialStatus: 'VERIFIED',
+    credentialPayload: 'string_representing_public_key_a',
+    credentialChallenge: 'string_representing_challenge_b'
+  }
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | Error {
```
eoln, line 62: let use ES Module (es6) syntax for exporting

seeds/02_scope.ts
```diff
@@ -0,0 +1,57 @@
...
+    accountId: '12345-67890'
+  },
+  {
+    consentId: '124',
+    action: 'accounts.transfer',
+    accountId: '21345-67890'
+  }
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | Error {
```
eoln, line 47: let use ES Module (es6) syntax for exporting

test/unit/cli.test.ts
```diff
@@ -33,11 +33,33 @@ describe('cli', (): void => {
       PACKAGE: Config.PACKAGE,
       PORT: Config.PORT,
       HOST: Config.HOST,
+      HOSTNAME: Config.HOSTNAME,
```
eoln, line 36: what is the difference between HOST and HOSTNAME ???

src/shared/config.ts
```diff
@@ -34,6 +34,31 @@ export interface ServiceConfig {
   PORT: number;
   HOST: string;
 
+  // ../lib/config.
+  HOSTNAME: string;
```
eoln, line 38: why HOSTNAME when HOST is already used?

test/unit/seeds/consent.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  afterAll(async (): Promise<void> => {
+    db.destroy()
+  })
+
+  it('should properly select all the entries in the Consent table', async (): Promise<void> => {
+    expect(db).toBeDefined()
+    const users: Knex.QueryBuilder[] = await db.from('Consent').select('*')
+    expect(users.length).toEqual(3)
+    expect(users[0]).toMatchObject({ id: '123', initiatorId: 'PISPA', participantId: 'DFSPA', credentialId: null, credentialType: null, credentialStatus: null, credentialPayload: null, credentialChallenge: null })
```
eoln, line 47: this part of code looks like magma, hard to read it in a fast way... Could you try reformatting, please?

migrations/20200624121732_consent.ts
```diff
@@ -0,0 +1,43 @@
...
+ ******/
+import * as Knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void | Knex.SchemaBuilder> => {
+  return knex.schema.createTableIfNotExists('Consent', (t: Knex.CreateTableBuilder): Knex.SchemaBuilder | void => {
+    t.string('id', 32).primary().notNullable()
+    t.string('initiatorId', 32).notNullable()
+    t.string('participantId', 32).notNullable()
+    t.timestamp('createdAt').defaultTo(knex.fn.now())
+    t.string('credentialId', 256).nullable()
```
ramanmangla, line 33: What is the `credentialId` length 256 based on?

test/unit/seeds/scope.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  })
+
+  afterAll(async (): Promise<void> => {
+    db.destroy()
+  })
+
+  it('should properly enforce the primary key constraint', async (): Promise<void> => {
+    expect(db).toBeDefined()
+    /* Tests for duplication */
+    await expect(db.from('Scope').insert({ id: 1, consentId: '125', action: 'accounts.transfer', accountId: '78901-12345' })).rejects.toMatchObject({
```
ramanmangla, line 69: Adding a `max-len: 80/100` linting parameter might be helpful in making such long lines more readable.

lewisdaly:
You will need to run `npm run audit:resolve` to try and fix the audit failures

seeds/01_consent.ts
```diff
@@ -0,0 +1,72 @@
...
+    credentialPayload: 'string_representing_public_key_a',
+    credentialChallenge: 'string_representing_challenge_b'
+  }
+]
+
+export function seed (knex: Knex): Knex.QueryBuilder<number[]> | Error {
+  try {
+    return knex('Consent').insert(consents)
+  } catch (err) {
+    if (err.code === 'ER_DUP_ENTRY') throw err
```
lewisdaly, line 66: So you are still throwing the error here but not logging to console?

migrations/20200624121732_consent.ts
```diff
@@ -38,6 +38,6 @@ exports.up = async (knex: Knex): Promise<void | Knex.SchemaBuilder> => {
   })
 }
 
-exports.down = function (knex: Knex): Knex.SchemaBuilder {
+export async function down (knex: Knex): Promise< void | Knex.SchemaBuilder> {
```
eoln, line 41: I was wrong the return value should be Knex.SchemaBuilder, not Promise: https://github.com/knex/knex/blob/master/types/index.d.ts#L1528

migrations/20200624121736_scope.ts
```diff
@@ -35,6 +35,6 @@ exports.up = async (knex: Knex): Promise<void | Knex.SchemaBuilder> => {
   })
 }
 
-exports.down = function (knex: Knex): Knex.SchemaBuilder {
+export async function down (knex: Knex): Promise< void | Knex.SchemaBuilder> {
```
eoln, line 38: I was wrong the return value should be Knex.SchemaBuilder, not Promise: https://github.com/knex/knex/blob/master/types/index.d.ts#L1528

sorry @spikerheado1234 :-)

migrations/migration.template
```diff
@@ -1,9 +1,9 @@
 import * as knex from 'knex'
 
-exports.up = async (knex: Knex): Promise<void> => {
+export async function up (knex: Knex): Promise<void | Knex.SchemaBuilder> {
 
 }
 
-exports.down = function(knex: Knex): Knex.SchemaBuilder {
+export async function down (knex: Knex): Promise< void | Knex.SchemaBuilder> {
```
eoln, line 7: I was wrong the return value should be Knex.SchemaBuilder, not Promise: https://github.com/knex/knex/blob/master/types/index.d.ts#L1528

migrations/20200624121732_consent.ts
```diff
@@ -24,7 +24,7 @@
  ******/
 import * as Knex from 'knex'
 
-exports.up = async (knex: Knex): Promise<void | Knex.SchemaBuilder> => {
+export async function up (knex: Knex): Promise<void | Knex.SchemaBuilder> {
   return knex.schema.createTableIfNotExists('Consent', (t: Knex.CreateTableBuilder): Knex.SchemaBuilder | void => {
     t.string('id', 32).primary().notNullable()
```
eoln, line 29: the return value from callback is only `void` now

migrations/20200624121736_scope.ts
```diff
@@ -24,7 +24,7 @@
  ******/
 import * as Knex from 'knex'
 
-exports.up = async (knex: Knex): Promise<void | Knex.SchemaBuilder> => {
+export async function up (knex: Knex): Promise<void | Knex.SchemaBuilder> {
```
eoln, line 27: the return value from callback is only `void` now

test/unit/cli.test.ts
```diff
@@ -38,6 +38,27 @@ describe('cli', (): void => {
         SHOW_HIDDEN: false,
         COLOR: true
       },
+      DATABASE: {
+        ACQUIRE_TIMEOUT_MILLIS: 30000,
+        CREATE_RETRY_INTERVAL_MILLIS: 200,
+        CREATE_TIMEOUT_MILLIS: 30000,
+        DEBUG: false,
+        DESTROY_TIMEOUT_MILLIS: 5000,
+        DIALECT: 'mysql',
```
jgeewax, line 47: Shouldn't this be `sqlite`? In other words, do tests still pass if you turn off your local MySQL database?

test/unit/seeds/consent.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      id: '123',
+      initiatorId: 'PISPA',
+      participantId: 'DFSPA',
+      credentialId: null,
+      credentialType: null,
+      credentialStatus: null,
+      credentialPayload: null,
+      credentialChallenge: null
+    })).rejects.toMatchObject({
+      code: 'SQLITE_CONSTRAINT',
```
jgeewax, line 106: Is this specific to SQLite? Is there a way to make sure that this is generic enough if someone changes the database engine/dialect?

test/unit/cli.test.ts
jgeewax: @spikerheado1234 : Thoughts?

PR #10, opened 2020-06-30
Feature/341: Add DB models and unit tests for Consent resource

ramanmangla:
In reference to ticket number #341. Addresses tasks 3 and 4.

This integrates the existing database migrations setup with the project by building data querying models on top of it.
- Implements querying models for creating, retrieving, updating and deleting Consent resources from the DB
- Adds unit tests using in memory SQLite to verify each model's functionality
- Adds cascades for the `consentId` foreign key in schema to update Scopes when the associated Consent is updated/deleted

Related future work will be addressed in following tickets:
 - `#355` Consistent Error Handling in Auth Service

Main Work Files:
 - `src/model/consent/consent.ts`
 - `src/model/consent/index.ts`
 - `test/unit/models/consent.test.ts`
 - `src/lib/db.ts`
 - `migrations/20200624121736_scope.ts`


migrations/20200624121732_consent.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Ahan Gupta <ahangupta.96@gmail.com>
+
+ --------------
+ ******/
+import * as Knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void> => {
```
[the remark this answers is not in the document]
eoln: Let use ES Module (es6) export syntax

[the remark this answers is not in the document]
eoln: try not to mix `function` and `=>` declarations in exports unless it is really necessary

migrations/20200624121736_scope.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Ahan Gupta <ahangupta.96@gmail.com>
+
+ --------------
+ ******/
+import * as Knex from 'knex'
+
+exports.up = async (knex: Knex): Promise<void> => {
```
eoln, line 27: Let use ES Module (es6) export syntax, use `function` instead ` =>`

seeds/01_consent.ts
```diff
@@ -0,0 +1,72 @@
...
+    participantId: 'DFSPA',
+    credentialId: '9875',
+    credentialType: 'FIDO',
+    credentialStatus: 'VERIFIED',
+    credentialPayload: 'string_representing_public_key_a',
+    credentialChallenge: 'string_representing_challenge_b'
+  }
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | number {
```
eoln, line 62: Let use ES Module (es6) export syntax

seeds/01_consent.ts
```diff
@@ -0,0 +1,72 @@
...
+  try {
+    return knex('Consent').insert(consents)
+  } catch (err) {
+    if (err.code === 'ER_DUP_ENTRY') return -1001
```
eoln, line 66: let keep the consistent way of returning the Errors from the `seed` function -> probably it will be fixed in @spikerheado1234 branch...

seeds/02_scope.ts
```diff
@@ -0,0 +1,60 @@
...
+  },
+  {
+    id: 3,
+    consentId: '124',
+    action: 'accounts.transfer',
+    accountId: '21345-67890'
+  }
+]
+
+exports.seed = function (knex: Knex): Knex.QueryBuilder<number[]> | number {
```
eoln, line 50: Let use ES Module (es6) export syntax

src/lib/config.ts
```diff
@@ -0,0 +1,68 @@
...
+ --------------
+ ******/
+
+import rc from 'rc'
+import Config from '../../config/default.json'
+
+const RC = rc('AS', Config)
+
+export const config = {
+  HOSTNAME: RC.HOSTNAME.replace(/\/$/, ''),
```
eoln, line 31: let use HOST variable which we already have in the main config file

src/model/consent/index.ts
```diff
@@ -0,0 +1,61 @@
...
+import Knex from 'knex'
+import {
+  Consent,
+  registerConsent,
+  updateCredentialsByConsentId,
+  getConsentById,
+  deleteConsentById
+} from './consent'
+
+export default class ConsentModel {
```
eoln, line 37: Let migrate `ConsentModel` class from src/model/consent/index to external file, move all db logic from functions in src/model/consent/consent to `ConsentModel` methods - there is no need to keep Object approach and Functional approach in the same time...

Simplify the names of methods to keep the code compact. There is no need to repeat `Consent` section in methods names. If you return Promise value the method should have `async` prefix

```
const consent = {...}
const consentModel = new ConsentModel(db)

// instead of 
consentModel.registerConsent(consent)

// use
consentModel.register(consent)
```

src/shared/config.ts
```diff
@@ -34,6 +34,31 @@ export interface ServiceConfig {
   PORT: number;
   HOST: string;
 
+  // ../lib/config.
+  HOSTNAME: string;
```
eoln, line 38: let use HOST instead of HOSTNAME

test/unit/models/consent.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  })
+
+  afterAll(async (): Promise<void> => {
+    Db.destroy()
+  })
+
+  describe('registerConsent', (): void => {
+    // Reset table for new test
+    beforeEach(async (): Promise<void> => {
+      await Db<Consent>('Consent').del()
```
eoln, line 70: Maybe instead of deleting the whole table, remove only rows added by tests. Jest tests are run in parallel and I am afraid that there is a risk that deleting the table could break other tests that are using it. Maybe this is a reason why sometimes your tests are failing randomly....

test/unit/models/consent.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      await Db<Consent>('Consent')
+        .insert(completeConsent)
+    })
+
+    it('deletes an existing consent', async (): Promise<void> => {
+      // Pre action Assertion
+      let users: Consent[] = await Db('Consent')
```
eoln, line 153: - The variable name `users` don't fit well. Maybe `consents` or `userConsents` would fit better?
- Use `const` instead of `let` if you are not reassigning the value.

test/unit/seeds/consent.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  afterAll(async (): Promise<void> => {
+    db.destroy()
+  })
+
+  it('should properly select all the entries in the Consent table', async (): Promise<void> => {
+    expect(db).toBeDefined()
+    const users: Knex.QueryBuilder[] = await db.from('Consent').select('*')
+    expect(users.length).toEqual(3)
+    expect(users[0]).toMatchObject({ id: '123', initiatorId: 'PISPA', participantId: 'DFSPA', credentialId: null, credentialType: null, credentialStatus: null, credentialPayload: null, credentialChallenge: null })
```
eoln, line 47: let reformat this code to be more readable

test/unit/seeds/scope.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  afterAll(async (): Promise<void> => {
+    db.destroy()
+  })
+
+  it('should properly select all the entries in the Scope table', async (): Promise<void> => {
+    expect(db).toBeDefined()
+    const users: Knex.QueryBuilder[] = await db.from('Scope').select('*')
+    expect(users.length).toEqual(3)
+    expect(users[0]).toEqual({ id: 1, consentId: '123', action: 'accounts.getBalance', accountId: '12345-67890' })
```
eoln, line 47: let reformat the code to be more readable

.eslintrc.js
```diff
@@ -33,6 +33,8 @@ module.exports = {
     'cucumber/expression-type': 2,
     'cucumber/no-restricted-tags': [2, 'wip', 'broken', 'foo'],
     'cucumber/no-arrow-functions': 2
+    // 'operator-linebreak': ["error", "after"],
+    // 'max-len': ["error", { "code": 100 }]
```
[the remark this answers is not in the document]
ramanmangla: I added in a max-len parameter but I was not sure whether it complied with Mojaloop's style guide (it doesn't exist in the other services). Should I include it again?

config/knexfile.ts
```diff
@@ -0,0 +1,39 @@
...
+      directory: seedsDirectory,
+      loadExtensions: ['.ts']
+    }
+  },
+  test: {
+    client: 'sqlite3',
+    connection: ':memory:',
+    useNullAsDefault: true,
+    migrations: {
+      directory: path.join(__dirname, '../migrations/'),
```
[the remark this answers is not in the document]
ramanmangla: Will be addressed in https://github.com/mojaloop/auth-service/pull/9

migrations/20200624121732_consent.ts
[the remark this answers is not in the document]
ramanmangla: Will be addressed in https://github.com/mojaloop/auth-service/pull/9

migrations/20200624121736_scope.ts
ramanmangla: Will be addresses in https://github.com/mojaloop/auth-service/pull/9

seeds/01_consent.ts
ramanmangla: Will be addresses in https://github.com/mojaloop/auth-service/pull/9

seeds/01_consent.ts
ramanmangla: Yes, it will be addressed in https://github.com/mojaloop/auth-service/pull/9. Our apologies for the mix up.

seeds/02_scope.ts
ramanmangla: Will be addressed in https://github.com/mojaloop/auth-service/pull/9

src/lib/config.ts
ramanmangla: Will be addressed in https://github.com/mojaloop/auth-service/pull/9

src/model/consent/index.ts
ramanmangla: Where do you suggest I move the ConsentModel class along with the logic? Do you have a new directory structure in mind?

src/shared/config.ts
ramanmangla: Will be addressed in https://github.com/mojaloop/auth-service/pull/9

test/unit/models/consent.test.ts
ramanmangla: For the time being, I am adding and deleting the same row for all the tests. So should I change the test data for each test as well? Otherwise, deleting only the row would mimic the same behaviour as deleting the entire table.

test/unit/seeds/consent.test.ts
ramanmangla: Will be addressed in https://github.com/mojaloop/auth-service/pull/9

test/unit/seeds/scope.test.ts
ramanmangla: Will be addressed in https://github.com/mojaloop/auth-service/pull/9

.eslintrc.js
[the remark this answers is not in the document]
eoln: I would say `yes` Do you agree with me @lewisdaly, @kleyow, @sridharvoruganti?

test/unit/models/consent.test.ts
eoln: keeping independent data between tests in the central database is a good practice and makes our life easy when working with tests.

ramanmangla: For synchronous tests, how can I remove specific rows anyways? I do not know which test will be executed next. Maybe I could delete the record in each test itself instead of using a beforeEach/AfterEach hook. What do you think?

test/unit/models/consent.test.ts
ramanmangla: Addressed in a recent commit

test/unit/models/consent.test.ts
eoln: If you will keep separate rows for each test then you can delete the whole table in `afterAll`... 
Maybe this article would give some hints ( or make things more complicated :-P ): 
https://walrus.ai/blog/2020/04/testing-database-interactions-with-jest/
Any ideas @lewisdaly @kleyow @sridharvoruganti ?

ramanmangla: Fair enough. I'll implement the changes.

ramanmangla: @eoln the walrus link that you shared talks about deleting all the rows in the DB using `beforeEach` hook for each test. So which approach should I follow? Deleting after every test or using different records for each test and deleting everything at the end of all tests?

src/model/consent/index.ts
ramanmangla: Addressed in a recent commit.

.eslintrc.js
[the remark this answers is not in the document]
lewisdaly: Sure go ahead.

test/unit/models/consent.test.ts
```diff
@@ -0,0 +1,207 @@
...
+        participantId: 'dfsp-3333-2123',
+        credentialId: null,
+        credentialType: null,
+        credentialStatus: null,
+        credentialPayload: null,
+        credentialChallenge: null
+      })
+    })
+
+    it('returns an error on adding a consent with existing consentId', async (): Promise<void> => {
```
lewisdaly, line 98: No need to have `Promise<void>` here... it's implied

src/model/consent/consent.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface Consent {
+  id: string;
+  initiatorId: string;
+  participantId: string;
+  credentialId?: number;
+  credentialType?: string;
+  credentialStatus?: string;
+  credentialPayload?: string;
+  credentialChallenge?: string;
+  timeStamp?: string;
```
lewisdaly, line 39: can this be `timestamp`?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,86 @@
...
+    return this.Db<Consent>('Consent')
+      .insert(consent)
+  }
+
+  // Update Consent credential
+  public updateCredentials (consent: Consent): Promise<Consent[]> {
+    return this.Db<Consent>('Consent')
+      .where({ id: consent.id })
+      .update({
+        credentialId: consent.credentialId,
```
lewisdaly, line 64: So my only concern here is that if you call `updateCredentials` with null or undefined fields, you could end up overwriting the existing fields.

Perhaps a better approach could be something like:
```js
const fieldsToUpdate = [ 'credentialId', 'credentialType', 'credentialStatus'...];
const updated = {}
Object.keys(consent).forEach(key => {
  if ( fieldsToUpdate.indexOf(key) > -1) {
    return;
  }
  const value = consents[key]
  if (!value) {
    return;
  }
  updated[key] = value
});

...then
.update(updated)
```

test/unit/models/consent.test.ts
```diff
@@ -0,0 +1,207 @@
+/*****
```
eoln, line 1: there should be `model` not `models` folder name -> folder name should be the same as in src

src/model/consent/consent.ts
eoln, line 39: as @jgeewax mentioned previously, timeStamp name could be confusing, it is better to use `createdAt`, `modifiedAt` etc., so the context of what kind of timestamp is clear.

test/unit/models/consent.test.ts
lewisdaly: The easy option is to separate everything into its own `describe` block.

The better approach is to keep track of which rows you have created, maybe make random ids, and delete them at the end of each test.

When you're creating models, add them to some list called `cleanup`

You can have a `beforeEach` that iterates through `cleanup` and deletes them 1 at a time.

src/model/consent/consent.ts
ramanmangla: Implemented changes in a recent commit.

test/unit/models/consent.test.ts
ramanmangla: Resolved in a recent commit.

src/model/consent/consent.ts
eoln: or better `createdAt` to give a context to timestamp?

test/unit/models/consent.test.ts
eoln: for me linter is complaining, maybe let update the rule?

src/model/consent/consent.ts
ramanmangla: Resolved in a recent commit.

test/unit/models/consent.test.ts
ramanmangla: But we are using `Promise<type>` as the return type for all async functions. Why not here?

ramanmangla: Yeah the linter wants return types for anonymous as well as named functions. Isn't that the convention though?

eoln: @lewisdaly would you agree to keep strictly typed return types? It looks little strange for some obvious cases, but we already have it used in so many places ;-)

src/model/consent/consent.ts
ramanmangla: Implemented in a recent commit.

src/model/consent/consent.ts
```diff
@@ -57,16 +66,20 @@ export class ConsentDB {
   }
 
   // Update Consent credential
+  // Only non-null Consent fields are updated
   public updateCredentials (consent: Consent): Promise<Consent[]> {
+    const validatedConsent = {}
+
+    for (const key of Object.keys(consent)) {
+      // Only allow nullProtectedFields with non-null values for update
+      if (this.nullProtectedProps.indexOf(key) !== -1 && Object(consent)[key]) {
```
eoln, line 75: is a need to use `Object(consent)[key]` instead of `consent[key]` ?

src/model/consent/consent.ts
```diff
@@ -57,16 +66,20 @@ export class ConsentDB {
   }
 
   // Update Consent credential
+  // Only non-null Consent fields are updated
   public updateCredentials (consent: Consent): Promise<Consent[]> {
+    const validatedConsent = {}
+
+    for (const key of Object.keys(consent)) {
```
eoln, line 73: you can use the simpler construct when apply `in` 
```
for (const property in object) {
  console.log(`${property}: ${object[property]}`);
}
```

src/model/consent/consent.ts
eoln: if nullProtectedProps will be the map: {credentialId: true, ... } than this code would be simpler and there will be no need to scan the whole array using indexOf

src/model/consent/consent.ts
```diff
@@ -45,9 +45,18 @@ export interface Consent {
 export class ConsentDB {
   // Knex instance
   private Db: Knex
+  // Nullable fields which don't allow for explicit null upsert
+  private nullProtectedProps: string[]
 
   public constructor (dbInstance: Knex) {
     this.Db = dbInstance
+    this.nullProtectedProps = [
```
eoln, line 53: `nullProtectedProps` should be a static property

src/model/consent/consent.ts
ramanmangla: Consent, being an interface, doesn't allow normal key indexing. Rest of your suggestions have been addressed in a recent commit.

src/model/consent/consent.ts
ramanmangla: Thank you for the suggestion! Implemented it.

src/model/consent/consent.ts
ramanmangla: Modified in a recent commit.

test/unit/models/consent.test.ts
lewisdaly: Ok no worries, yeah I just thought it looked strange, no major qualms here.

src/model/consent/consent.ts
```diff
@@ -1,3 +1,5 @@
+/* istanbul ignore file */
```
eoln, line 1: why this line?

ramanmangla: @lewisdaly suggested adding this because we haven't implemented BDD tests for models. These will be addressed in future tickets.

ramanmangla: So is there another workaround for now?

lewisdaly: I think it's fine we just need to also have a comment here that explains why we are ignoring it _for now_, that references a ticket for implementing the BDD tests for end to end consents creation/updating

ramanmangla: Addressed in a recent commit.

src/model/consent/consent.ts
```diff
@@ -0,0 +1,105 @@
...
+    const validatedConsent = {}
+
+    for (const key in consent) {
+      // Only allow nullProtectedFields with non-null values for update
+      if (ConsentDB.nullProtectedProps.get(key) && Object(consent)[key]) {
+        Object(validatedConsent)[key] = Object(consent)[key]
+      }
+    }
+
+    return this.Db<Consent>('Consent')
```
eoln, line 86: @lewisdaly @jgeewax there is a corresponding test which ensures we do nothing (no update) if there is no related Consent in DB. Shouldn't we throw an error in such a case?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,105 @@
+/* istanbul ignore file */
+
+/*
+ * This flag is to ignore BDD testing for model
+ * which will be addressed in a future ticket
```
lewisdaly, line 5: Please open ticket and reference explicitly here

src/model/consent/consent.ts
lewisdaly: Hmm probably yes. I didn't notice this earlier. I think I assumed that doing a `.where()` in knex when the id doesn't exist would throw an error.

ramanmangla: So my thought here was that the models were just supposed to act as wrappers over the queries and the actual validation was supposed to happen in domains. Just like SQL, Knex doesn't return an error when it doesn't find a specific row to update. Instead, it does nothing and returns. So I was thinking of checking whether the consent actually exists in the domain logic before even trying to update. Any thoughts? @lewisdaly @jgeewax @eoln

ramanmangla: This is the case with all the models. I'm adhering to the way SQL would work and letting domains handle the validation.

jgeewax: OK I think this is a bigger issue of where validation happens. Can you tell me now what happens if you attempt to update a resource that doesn't exist? Will it just fail silently? Throw an error? Create the row that would've been missing?

If Mojaloop has a standard of what we do here, we should follow that. If not, then we should be explicit about what is and is not permitted. Keep in mind here that putting this inside a controller (rather than at a model level) is likely to be problematic unless Knex provides a way to manage transactional semantics.

In other words, if a handler has to do the following, then we have a problem:

1. Check if item exists. If not, return 404.
2. Update item

What if the item exists but is deleted between steps (1) and (2)? What if the item doesn't exist before step (1) but then exists right after?

The second scenario isn't that scary (just retry the request). The first is very scary if our code has "undefined behavior". As a result, we **must** define the behavior of what happens when someone tries to update a row that doesn't exist. And that behavior (in my opinion) should be to throw some sort of error / reject a promise.

jgeewax: OK I just checked the [docs](http://knexjs.org/#Builder-update) more thoroughly. If nothing matches, no updates happen and the result includes an empty array.

In other words, we decide whether a failure happened by looking at the result -- if it's empty, that's a 404 Not Found response code. That's a bit unusual, but not a big deal and avoids transactions entirely which is always fun.

src/model/consent/consent.ts
ramanmangla: I have added #354 ticket with basic goal on the PISP board. Please let me know if that works. Do you want me to just mention the ticket number here?

src/model/consent/consent.ts
ramanmangla: We definitely need error handling and validation for checking whether a row exists or not. I thought we could implement it in domains but I agree that it's a better idea to have validation in the model itself. For now, I can return a 404 error from the model but consistent error handling and edge cases will be handled in a future error handling ticket as we have discussed earlier.

src/model/consent/index.ts
```diff
@@ -0,0 +1,28 @@
+/*****
+ License
+ --------------
+ Copyright © 2017 Bill & Melinda Gates Foundation
```
eoln, line 4: 2020

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,249 @@
+/*****
+ License
+ --------------
+ Copyright © 2017 Bill & Melinda Gates Foundation
```
eoln, line 4: 2020

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,249 @@
...
+      await consentDB.register(partialConsent)
+
+      // Assertion
+      const consents: Consent[] = await Db<Consent>('Consent')
+        .select('*')
+        .where({
+          id: partialConsent.id
+        })
+
+      expect(consents[0].createdAt).toEqual(expect.anything())
```
eoln, line 85: you can use `createdAt: expect.anything()` in L87, so L85 can be dropped

eoln: for other cases, you can apply it also

src/lib/db.ts
```diff
@@ -0,0 +1,35 @@
+/*****
+ License
+ --------------
+ Copyright © 2017 Bill & Melinda Gates Foundation
```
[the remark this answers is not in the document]
jgeewax: Mojaloop Foundation* please.

src/model/consent/consent.ts
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+/*****
+ License
+ --------------
+ Copyright © 2017 Bill & Melinda Gates Foundation
```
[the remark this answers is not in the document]
jgeewax: Mojaloop Foundation*

src/model/consent/consent.ts
```diff
@@ -0,0 +1,113 @@
...
+export interface Consent {
+  id: string;
+  initiatorId: string;
+  participantId: string;
+  credentialId?: string;
+  credentialType?: string;
+  credentialStatus?: string;
+  credentialPayload?: string;
+  credentialChallenge?: string;
+  createdAt?: string;
```
jgeewax, line 46: Do we not want to use a time type of some sort here?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import Knex from 'knex'
+
+/*
+ * Interface for Consent resource type
+ */
+export interface Consent {
+  id: string;
+  initiatorId: string;
+  participantId: string;
```
jgeewax, line 40: Are we sure these are required? Won't be want to do a lookup by ID and need these to be empty during that lookup or no?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,113 @@
...
+    ['credentialChallenge', true]
+  ])
+
+  public constructor (dbInstance: Knex) {
+    this.Db = dbInstance
+  }
+
+  // Add initial Consent parameters
+  public register (consent: Consent): Promise<void> {
+    return this.Db<Consent>('Consent')
```
jgeewax, line 70: Isn't this a promise for an integer? I thought `.insert()` returns the number of rows inserted, no?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    const numUpdated: number = await this
+      .Db<Consent>('Consent')
+      .where({ id: consent.id })
+      .update(validatedConsent)
+
+    // Ensure that the caller knows that the resource does not exist
+    if (numUpdated === 0) {
+      throw Error('Error 404: Consent ID does not exist')
```
jgeewax, line 93: Yikes -- this was definitely not what I had intended... It's fine here to let this method return a zero value. (HTTP errors shouldn't ever trickle down to this abstraction layer.)

Instead, we should probably just return what's on like 86 (`return this.Db<Consent> ....`)

src/model/consent/consent.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    return numUpdated
+  }
+
+  // Retrieve Consent by ID
+  public retrieve (id: string): Promise<Consent[]> {
```
jgeewax, line 100: Shouldn't we have a limit of 1 and always return a single matching consent or a null value? So this would be a Promise<Consent> returned and we'd include a limit() statement in the query?

src/model/consent/index.ts
jgeewax: Mojaloop Foundation*

src/model/consent/consent.ts
ramanmangla: I agree that the return type of a Promise containing a single Consent/null would be better. I'll implement that. 

I didn't include a limit of 1 consent here because the DB cannot have replicated consentIDs in the first place. Should I still have a limit?

src/model/consent/consent.ts
ramanmangla: The fields are `required` to enforce that they are a part of each request that the auth-service receives (which is the case in the sequence diagrams). The `retrieve` consent model just requires a `consentID` (String) for lookup and doesn't concern the interface. Do you feel that's justified?

src/model/consent/consent.ts
ramanmangla: Yeah, I thought we didn't require it but we can have it for validation purposes.

src/model/consent/consent.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/*
+ * Class to abstract Consent DB operations
+ */
+export class ConsentDB {
+  // Knex instance
+  private Db: Knex
+  // Nullable fields which don't allow for explicit null upsert
+  private static nullProtectedProps: Map<string, boolean> = new Map([
```
lewisdaly, line 61: I always thought `{}` was faster than using a js `Map` object, but I'm happy to be corrected.

src/model/consent/consent.ts
lewisdaly: Yeah, put the limit in, and just return the `Consent`. Keep in mind that those assumptions may change in the future (even though that's very unlikely), so putting them in the code explicitly here is a way to make sure our code breaks when our assumptions change.

src/model/consent/consent.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    const numUpdated: number = await this
+      .Db<Consent>('Consent')
+      .where({ id: consent.id })
+      .update(validatedConsent)
+
+    // Ensure that the caller knows that the resource does not exist
+    if (numUpdated === 0) {
+      throw Error('Error 404: Consent ID does not exist')
```
jgeewax, line 107:
```suggestion
      .Db<Consent>('Consent')
      .where({ id: consent.id })
      .update(validatedConsent)

    // Ensure that the caller knows that the resource does not exist
    if (numUpdated === 0) {
      throw NotFoundError("Consent", consent.id)
```

jgeewax: Also somewhere in a separate file...

We can update the constructor here to take both a resource name (e.g., "Consent") and an ID (e.g., 1) and auto-generate the error string.

```typescript
class NotFoundError extends Error {
}
```

src/model/consent/consent.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    return numUpdated
+  }
+
+  // Retrieve Consent by ID
+  public retrieve (id: string): Promise<Consent[]> {
+    return this.Db<Consent>('Consent')
+      .select('*')
+      .where({ id: id })
```
jgeewax, line 117:
```suggestion
    let result = await this.Db<Consent>('Consent')
      .select('*')
      .where({ id: id })
    if (result.length == 0)
      throw NotFoundError(Consent, id)
    return result[0];
```

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  describe('register', (): void => {
+    // Reset table for new test
+    beforeEach(async (): Promise<void> => {
+      await Db<Consent>('Consent').del()
+    })
+
+    it('adds consent with partial info to the database', async (): Promise<void> => {
+      // Action
+      const numInserted: number = await consentDB.register(partialConsent)
```
jgeewax, line 81:
```suggestion
      const insertCount: number = await consentDB.register(partialConsent)
```

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      // Fail primary key constraint
+      await expect(consentDB.register(partialConsent))
+        .rejects.toThrowError()
+    })
+  })
+
+  describe('updateCredentials', (): void => {
+    // Reset table for new test
+    beforeEach(async (): Promise<void> => {
+      await Db<Consent>('Consent').del()
```
jgeewax, line 138: There should be a way for SQLite should auto-clear the data automatically. But have to double check....

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    })
+
+    it('updates credentials for existing consent', async (): Promise<void> => {
+      // Inserting record to update
+      await Db<Consent>('Consent')
+        .insert(partialConsent)
+
```
jgeewax, line 145:
```suggestion
expect(partialConsent.id).toEqual(completeConsent.id)
```

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      await consentDB.updateCredentials(completeConsent)
+
+      // Assertion
+      const consents: Consent[] = await Db<Consent>('Consent')
+        .select('*')
+        .where({
+          id: completeConsent.id
+        })
+
+      expect(consents[0].createdAt).toEqual(expect.anything())
```
jgeewax, line 156: Can we check that this is an instance of `Date` ?

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,254 @@
...
+
+      expect(consents[0]).toEqual(expect.objectContaining(completeConsent))
+    })
+
+    it('throws an error for a non-existent consent', async (): Promise<void> => {
+      // Action
+      await expect(consentDB.updateCredentials(completeConsent))
+        .rejects.toThrowError('Error 404: Consent ID does not exist')
```
jgeewax, line 164: Need to update this one.

eoln:
discussion on how to use Date & Knex
https://stackoverflow.com/questions/41916012/storing-a-node-js-date-in-a-knex-datetime-mysql-datetime-field

test/unit/model/consent.test.ts
ramanmangla: Design decision required about MySQL timestamp return and implementation of Data methods.

test/unit/model/consent.test.ts
ramanmangla: We can destroy the DB connection and start up a new connection for each test but that would take more time than just clearing the tables. I haven't been able to find an other information on this.

src/model/error.ts
```diff
@@ -0,0 +1,35 @@
...
+ their names indented and be marked with a '-'. Email address can be added
+ optionally within square brackets <email>.
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Raman Mangla <ramanmangla@google.com>
+ --------------
+ ******/
+
+export default class NotFoundError extends Error {
```
eoln, line 30:
```suggestion
export class NotFoundError extends Error {
```
there will be more errors, rename the filename to `errors.ts`

src/model/consent/consent.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  // Retrieve Consent by ID (unique)
+  public async retrieve (id: string): Promise<Consent[]> {
+    // Returns array containing consents
+    const consents: Consent[] = await this
+      .Db<Consent>('Consent')
+      .select('*')
+      .where({ id: id })
+      .limit(1)
```
akapur99, line 124: maybe i'm just misunderstanding but 1) if ID is unique then then why are you returning an array? 2) if you're returning an array what does the `limit(1)` do?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,141 @@
...
+    // Ensure that the caller knows that the resource does not exist
+    if (updateCount === 0) {
+      throw new NotFoundError('Consent', consent.id)
+    }
+
+    return updateCount
+  }
+
+  // Retrieve Consent by ID (unique)
+  public async retrieve (id: string): Promise<Consent[]> {
```
akapur99, line 118: maybe i'm just misunderstanding but 1) if ID is unique then then why are you returning an array? 2) if you're returning an array what does the limit(1) do?

ramanmangla: Thanks for pointing that out!

src/model/consent/consent.ts
jgeewax: Sounds like the DFSP is responsible for specifying the Consent ID in the POST (create) request. So this should be required for sure.

src/model/consent/consent.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  // Update Consent credential
+  // Only non-null Consent fields are updated
+  public async updateCredentials (consent: Consent): Promise<number> {
+    const validatedConsent = {}
+
+    for (const key in consent) {
+      // Only allow nullProtectedFields with non-null values for update
+      if (ConsentDB.nullProtectedProps[key] && (consent as unknown as Record<string, boolean>)[key]) {
```
jgeewax, line 98: Completely out of my depth, but what about:

`consent[key as keyof typeof consent]`

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html

?

Also, we should also check for explicit nulls or undefined. Because 0 values (not null, but 0 integers) would throw an error here when they shouldn't! Or look at hapi joi for validation? Or maybe just don't do this at all here and leave this for the handler layer to address?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    return insertCount[0]
+  }
+
+  // Update Consent credential
+  // Only non-null Consent fields are updated
+  public async updateCredentials (consent: Consent): Promise<number> {
+    const validatedConsent = {}
+
+    for (const key in consent) {
```
jgeewax, line 96: If I have an interface `Thing` with properties `a: string` and `b: string?`, and I create this interface as `Thing({ a: "a" })`, when I iterate, will I see `key` values of both `"a"` and `"b"` ? Or just `"a"`? If just a, is that OK?

eoln: in TS you can't have an instance of an interface, but only the object which conforms to interface declarations. So the answer is: for object {a: 'a'} which conforms to Thing interface you will have just `a`.

src/model/consent/consent.ts
```diff
@@ -0,0 +1,148 @@
+/* istanbul ignore file */
+
+/*
```
ramanmangla, line 3: Add ticket # addressing BDD here.

src/model/consent/consent.ts
jgeewax, line 98: `consent[key as keyof Consent]`

PR #11, opened 2020-07-02
#339 Handling Challenge Generation Request

akapur99:
In reference to Ticket #339 - Deals with : 
1. Request handler for POST /consents/{id}/generateChallenge
2. Validation of inbound request
3. Generation of the challenge (following FIDO standards)
4. Issue outbound call to PISP (create requests library)
5. Unit Testing for all of the above

Also includes method to convert ExternalScopes to Scopes and testing for the same. 
Integration testing will be handled in a separate PR.


src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,65 @@
...
+    // Challenge generation
+    let challenge = ''
+    Crypto.randomBytes(32, (err: Error, buf): void => {
+      if (err) throw err
+      challenge = buf.toString('base64')
+    })
+
+    // Update consent credentials
+    consent.credentialType = 'FIDO'
+    consent.credentialStatus = 'ACTIVE'
```
ramanmangla, line 56: The `credentialStatus` in `generateChallenge` is 'PENDING'.

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,76 @@
...
+import { randomBytes } from 'crypto'
+const Enum = require('@mojaloop/central-services-shared').Enum
+
+export const isConsentRequestValid = function (request: Request, consent: Consent): boolean {
+  const fspiopSource = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
+
+  return (consent && consent.initiatorId === fspiopSource)
+}
+
+export const genChallenge = async function (request: Request, consent: Consent): Promise<void> {
```
eoln, line 40: I suggest renaming this function to `generateChallenge`. You should skip declaring a const variable with anonymous function as a value (it is antipattern). Instead, export named function.

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (consent.credentialChallenge) {
+    try {
+      putConsentId(consent, request.headers)
+    } catch (error) {
+      console.warn(error)
+    }
+    return
+  }
+
+  // Challenge generation
```
eoln, line 52: you can put challenge string generation logic into an async separate function (separation of the concerns). It will be much easier to test it without the burden of consentDB.updateCredentials

src/server/domain/consents/{ID}/generateChallenge.ts
eoln: please add documentation for this function -> the description where you explain what for it should be used, what logical steps are done (generation of the challenge value, store it, update the client about it via async PUT

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  return false
+}
+
+export const genChallenge = async function (request: Request, consent: Consent): Promise<void> {
+  // If there is no pre-existing challenge for the consent id
+  // Generate one and update the database
+  if (consent.credentialChallenge == null) {
+    // Challenge generation
+    const crypto = require('crypto')
```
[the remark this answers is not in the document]
akapur99: yup just refactored this in the morning - this commit is outdated

akapur99:
Just rebased on master - to incorporate changes made in @spikerheado1234 's PR

src/server/domain/consents/{ID}/generateChallenge.ts
spikerheado1234: This is being a bit nitpicky here but the function is named `generateChallenge` but what you do is you generate the challenge, followed by storing it, following by updating the client via a PUT. Thus the name `generateChallenge` is not as apt as it should be.

Additionally, I feel like we should separate the concerns apart. In `generateChallenge` should be code to solely generate a challenge, whereas both the outgoing `PUT` request as well as the storing into the db should be abstracted into two other functions and called accordingly.

akapur99: Agree with the thoughts on function naming and modularization. the putConsentID is actually a separate function already which is building the body and then making the PUT call. Will refactor the rest. Thought I think its useful to leave the main function name as generateChallenge to signify it as the main function with the domain  file?

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,112 @@
...
+ * a secure random challenge string (Base 64 encoding)
+ */
+export async function generateChallengeValue (): Promise<string> {
+  const randBytes = promisify(randomBytes)
+  try {
+    const buf = await randBytes(32)
+    const challenge = buf.toString('base64')
+    return challenge
+  } catch (error) {
+    throw new Error('Error in Generating Challenge Value')
```
lewisdaly, line 59: Is there any reason to not just throw the original error?

src/shared/requests.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export async function putConsentId (consent: Consent, headers): Promise<void> {
+  // Switch SOURCE and DESTINATION in headers
+  const destinationId = headers[Enum.Http.Headers.FSPIOP.SOURCE]
+  headers[Enum.Http.Headers.FSPIOP.SOURCE] = headers[Enum.Http.Headers.FSPIOP.DESTINATION]
+  headers[Enum.Http.Headers.FSPIOP.DESTINATION] = destinationId
+
+  // Retrieve scopes
+  let scopes: Scope = null
+  scopes = await scopeDb.retrieve(consent.id)
```
lewisdaly, line 17: Can this just be `const scopes = await scopeDb.retrieve(consent.id)`?

lewisdaly:
Hmm. I'm not sure what's going on with CircleCi here...

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,113 @@
...
+const Enum = require('@mojaloop/central-services-shared').Enum
+
+/**
+ * Validates whether generate challenge request is valid
+ * by comparing consent ID sent matches with existing consent in table
+ * and if source ID matches initiator ID
+ * @param request: request received from PISP
+ * @param consent: Consent object
+ */
+export const isConsentRequestValid = function (request: Request, consent: Consent): boolean {
```
jgeewax, line 42: Let's use https://github.com/hapijs/joi as the rest of Mojaloop does.

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,113 @@
...
+
+ --------------
+ ******/
+
+import { Request } from '@hapi/hapi'
+import { consentDB } from '../../../../lib/db'
+import { Consent } from '../../../../model/consent'
+import { putConsentId } from '../../../../shared/requests'
+import { promisify } from 'util'
+import { randomBytes } from 'crypto'
```
jgeewax, line 31:
```typescript
import * from 'crypto';
const randBytes = promisify(crypto.randBytes)
```

This is not perfect, you get the idea though... :-)

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  return (consent && consent.initiatorId === fspiopSource)
+}
+
+/**
+ * Helper function which uses the crypto library to generate
+ * a secure random challenge string (Base 64 encoding)
+ */
+export async function generateChallengeValue (): Promise<string> {
+  const randBytes = promisify(randomBytes)
```
jgeewax, line 53: Move this out as per previous comment (for perf reasons)

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,113 @@
...
+  try {
+    const buf = await randBytes(32)
+    const challenge = buf.toString('base64')
+    return challenge
```
jgeewax, line 57:
```suggestion
    return buf.toString('base64')
```

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,113 @@
...
+  const fspiopSource = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
+
+  return (consent && consent.initiatorId === fspiopSource)
+}
+
+/**
+ * Helper function which uses the crypto library to generate
+ * a secure random challenge string (Base 64 encoding)
+ */
+export async function generateChallengeValue (): Promise<string> {
```
jgeewax, line 52: This function might belong in a different file as it's useful outside the context of the HTTP routing.

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  // If consent is invalid, throw error
+  if (!isConsentRequestValid(request, consent)) {
+    throw new Error('400')
+  }
+
+  // Asynchronously generate challenge
+  try {
+    generateChallenge(request, consent)
```
jgeewax, line 56:
```suggestion
    process.nextTick( async () => {
        let challengeValue = await generateChallengeValue()  // randomBytes...
        let result = await updateConsent(..., challengeValue)
        returnCallViaPut(result)
    })
```

jgeewax: Please also add some error handling in case something goes wrong... (Just a TODO)

src/shared/requests.ts
```diff
@@ -0,0 +1,66 @@
...
+      credentialStatus: consent.credentialStatus,
+      challenge: {
+        payload: consent.credentialChallenge,
+        signature: null
+      },
+      payload: null
+    }
+  }
+  // Use sdk-standard-components library to send request
+  putConsents(body, destinationId, consent.id, headers)
```
jgeewax, line 63:
```suggestion
  return putConsents(body, destinationId, consent.id, headers)
```

src/server/handlers/consents/{ID}/generateChallenge.ts
eoln: I was wrong about the semantic of `process.nextTick(..)` it has the potential to be dangerous and should be used very carefully. The `process.setImmediate(..)` should be used instead: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/

src/server/domain/consents/{ID}/generateChallenge.ts
akapur99: @jgeewax just spoke with @eoln - joi doesn't seem to make sense in this situation, plus if we validate by joi it seems we'd have to add validation for all keys in the request.parameters + request.headers (based on what I just read in the joi docs) 
IIIRC your main concern was adding codes / specific error messages for different failing conditions? in that case could we implement that in the function throwing errors or returning the corresponding response code instead of the '202' code ?

jgeewax: OK - and this is because the validation we're doing here isn't about the schema and format of the inputs, but instead that specific fields line up with expected environment parameters? If so, that's fine. Just want to make sure we're clear on what we're validating.

akapur99: yes, essentially we just want to check that the consent ID sent to us - correlates to a Consent object in our database AND that the initiator ID of the request matches with the initiator ID stored in the consent object.

eoln: @jgeewax we are using `hapi-openapi` plugin which generates input validation from the OpenApi schema document: https://github.com/krakenjs/hapi-openapi#readme

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (!isConsentRequestValid(request, consent)) {
+    throw new Error('400')
+  }
+
+  // Asynchronously deals with generating challenge, updating consent db
+  //  and making outgoing PUT consent/{ID} call
+
+  // remove try catch
+
+  try {
```
eoln, line 61: there is no need for this try-catch

test/unit/server/domain/consents/{ID}/generateChallenge.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/*
+ * Mock Consent Resources
+ */
+const partialConsent: Consent = {
+  id: '1234',
+  initiatorId: 'pisp-2342-2233',
+  participantId: 'dfsp-3333-2123'
+}
+
```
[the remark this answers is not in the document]
akapur99: there's a slight difference wrt the initiator id - which is being used to test error handling :)

src/shared/requests.ts
```diff
@@ -0,0 +1,91 @@
...
+  const destinationId = headers[Enum.Http.Headers.FSPIOP.SOURCE]
+  // eslint-disable-next-line max-len
+  headers[Enum.Http.Headers.FSPIOP.SOURCE] = headers[Enum.Http.Headers.FSPIOP.DESTINATION]
+  headers[Enum.Http.Headers.FSPIOP.DESTINATION] = destinationId
+
+  // Retrieve scopes
+  const scopesRetrieved: Scope[] = await scopeDb.retrieve(consent.id)
+
+  // Reformat scopes to match what external handler wants
+  const temp = {}
```
jgeewax, line 55:
```suggestion
  const ANYTHING_ELSE = {}
```

src/shared/requests.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  scopesRetrieved.forEach((scope: Scope): void => {
```
jgeewax, line 57: Why not for of?

src/shared/requests.ts
```diff
@@ -0,0 +1,91 @@
...
+  // Switch SOURCE and DESTINATION in headers
+  const destinationId = headers[Enum.Http.Headers.FSPIOP.SOURCE]
+  // eslint-disable-next-line max-len
+  headers[Enum.Http.Headers.FSPIOP.SOURCE] = headers[Enum.Http.Headers.FSPIOP.DESTINATION]
+  headers[Enum.Http.Headers.FSPIOP.DESTINATION] = destinationId
+
+  // Retrieve scopes
+  const scopesRetrieved: Scope[] = await scopeDb.retrieve(consent.id)
+
+  // Reformat scopes to match what external handler wants
```
jgeewax, line 54: This is taking a list of actions and putting it into an array attached to the consent.

jgeewax, line 54: This might benefit from a separate function...

src/shared/requests.ts
```diff
@@ -0,0 +1,93 @@
...
+      temp[accountId] = {
+        id: scope.id,
+        consentId: consent.id,
+        accountId,
+        actions: [scope.action]
+      }
+    }
+  })
+
+  const scopes: Scope[] = Object.values(temp)
```
akapur99: write a shared function - mapper
Write a TS definition 
src/shared/newFile

src/shared/requests.ts
```diff
@@ -0,0 +1,93 @@
...
+ their names indented and be marked with a '-'. Email address can be added
+ optionally within square brackets <email>.
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+
+ --------------
+ ******/
+import { putConsents } from '@mojaloop/sdk-standard-components'
```
akapur99, line 30: MIGRATE entire function / module stuff to domain logic

src/lib/db.ts
```diff
@@ -30,10 +30,38 @@
...
+import {ScopeDB, Scope} from '../model/scope'
 const Db: Knex = Knex(Config.test)
 const consentDB: ConsentDB = new ConsentDB(Db)
 
+/**
+ * Retrieves scopes from database, reformats object structure, removing
+ * scope & consent ids, and returns array of scopes
+ * @param id Consent id
+ */
+export async function retrieveScopes (id: string): Promise<Scope[]> {
```
ramanmangla, line 42: Do we need this function here? Seems like domain logic for this specific handler. Also, this file was just to make the DB connection.

src/lib/db.ts
```diff
@@ -30,10 +30,38 @@
...
 import Config from '../../config/knexfile'
 import ConsentDB from '../model/consent'
-
+import {ScopeDB, Scope} from '../model/scope'
 const Db: Knex = Knex(Config.test)
 const consentDB: ConsentDB = new ConsentDB(Db)
 
+/**
+ * Retrieves scopes from database, reformats object structure, removing
+ * scope & consent ids, and returns array of scopes
```
ramanmangla, line 39: The return type in docs doesn't match the actual return type.

src/lib/db.ts
eoln: It is domain logic, let move it there. there is a need to rethink the usage and return type, where this

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+
+ --------------
+ ******/
+// eslint-disable-next-line max-len
+import { updateCredential, isConsentRequestValid, putConsentId } from '../../../domain/consents/{ID}/generateChallenge'
+import { generate } from '../../../../lib/challenge'
```
eoln, line 32:
```suggestion
import { * as challenge } from '../../../../lib/challenge'
```

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  // Asynchronously deals with generating challenge, updating consent db
+  //  and making outgoing PUT consent/{ID} call
+  setImmediate(async (): Promise<void> => {
+    try {
+      // If there is no pre-existing challenge for the consent id
+      // Generate one and update database
+      if (!consent.credentialChallenge) {
+        // Challenge generation
+        const challenge = await generate()
```
eoln, line 71:
```suggestion
        const challengeValue = await challenge.generate()
```

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        // Updating credentials with generated challenge
+        consent = await updateCredential(consent, challenge, 'FIDO', 'PENDING')
```
eoln, line 74:
```suggestion
        consent = await updateCredential(consent, challengeValue, 'FIDO', 'PENDING')
```

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,100 @@
...
+  consent: Consent): boolean {
+  const fspiopSource = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
+  return (consent && consent.initiatorId === fspiopSource)
+}
+
+/**
+ * Assigns credentials to given consent object and updates in the database
+ * Returns updated consent object
+ */
+export async function updateCredential (
```
eoln, line 56:
```suggestion
export async function updateConsentCredential (
```

eoln: this code looks for me as a good candidate to be migrated to the ConsentDB model as a method `updateCredential(consent: Consent, credential: ConsentCredential)`, It would be wonderful if you could introduce a new ConsentCredential interface containing these three properties: challenge, status, type

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,100 @@
...
+  challenge: string,
+  credentialType: string,
+  credentialStatus: string): Promise<Consent> {
+  // Update consent credentials
+  consent.credentialType = credentialType
+  consent.credentialStatus = credentialStatus
+  consent.credentialChallenge = challenge
+
+  // Update in database
+  await consentDB.update(consent)
```
eoln, line 67:
```suggestion
   await.consentDB.updateCredential( consent, credential }
```

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,100 @@
...
+  return (consent && consent.initiatorId === fspiopSource)
+}
+
+/**
+ * Assigns credentials to given consent object and updates in the database
+ * Returns updated consent object
+ */
+export async function updateCredential (
+  consent: Consent,
+  challenge: string,
```
eoln, line 58: let use here `ConsentCredential` interface instead of multiple arguments

src/server/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,100 @@
...
+  request: Request):
+  Promise<SDKStandardComponents.GenericRequestResponse | undefined> {
+  // Construct body of outgoing request
+  const body = {
+    requestId: consent.id,
+    initiatorId: consent.initiatorId as string,
+    participantId: consent.participantId as string,
+    scopes,
+    credential: {
+      id: null,
```
eoln, line 87: why `null` here

test/unit/lib/db.test.ts
```diff
@@ -0,0 +1,77 @@
...
+ their names indented and be marked with a '-'. Email address can be added
+ optionally within square brackets <email>.
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+
+ --------------
+ ******/
+
```
eoln, line 30: I think this mocking isn't working properly: https://jestjs.io/docs/en/mock-functions#mock-implementations

src/server/domain/consents/{ID}/generateChallenge.ts
ramanmangla: @eoln we actually tried many variations of `update` for `Consents` and decided on the correct one. The current one can take in any of the fields that you pass in and updates only those. This is because different endpoints update different parts of the `Consent` resource` and it wouldn't be feasible to have 3-4 variations of `update` for these things. Instead, we choose to stay usage agnostic in the model and let handler decide what to update.

src/server/domain/consents/{ID}/generateChallenge.ts
akapur99: this is based on the linking diagram

akapur99: https://camo.githubusercontent.com/57f9a75d95d5bdd24f7563ce5c81579c8aa9db24/687474703a2f2f7777772e706c616e74756d6c2e636f6d2f706c616e74756d6c2f70726f78793f63616368653d6e6f267372633d68747470733a2f2f7261772e67697468756275736572636f6e74656e742e636f6d2f6d6f6a616c6f6f702f706973702f6d61737465722f646f63732f6c696e6b696e672f35612d63726564656e7469616c2d726567697374726174696f6e2e70756d6c

test/unit/lib/db.test.ts
eoln: @ramanmangla could you pair with @akapur99 to share the mocking knowledge?

src/lib/scopes.ts
```diff
@@ -50,11 +50,13 @@ export function convertScopesToExternal (
     const accountId: string = scope.accountId
 
     if (!(accountId in scopeDictionary)) {
+      // @ts-ignore
```
eoln, line 53: `ts-ignore` should be banned :-)

test/unit/server/domain/consents/{ID}/generateChallenge.test.ts
```diff
@@ -28,18 +28,21 @@
  --------------
  ******/
 import { Request } from '@hapi/hapi'
-import { consentDB, retrieveScopes } from '../../../../../../src/lib/db'
+import { consentDB } from '../../../../../../src/lib/db'
 import { Consent } from '../../../../../../src/model/consent'
 import { thirdPartyRequest } from '../../../../../../src/lib/requests'
 // eslint-disable-next-line max-len
-import { updateCredential, isConsentRequestValid, putConsentId } from '../../../../../../src/server/domain/consents/{ID}/generateChallenge'
+import { updateConsentCredential, isConsentRequestInitiatedByValidSource, putConsentId } from '../../../../../../src/server/domain/consents/{ID}/generateChallenge'
```
eoln, line 35:
```suggestion
import {
  updateConsentCredential,
  isConsentRequestInitiatedByValidSource,
  putConsentId 
} from '../../../../../../src/server/domain/consents/{ID}/generateChallenge'
```

eoln, line 35: try to reformat before disabling linting requirements

src/lib/scopes.ts
```diff
@@ -50,11 +50,13 @@ export function convertScopesToExternal (
     const accountId: string = scope.accountId
```
eoln, line 51: what for `ExternalScope` is introduced ?

eoln:
@akapur99 please move `src/server/domain` folder to `src/domain`. In `src/server` folder are only files related to HTTP server setup

src/lib/scopes.ts
akapur99: there's a mismatch between the scope interface of the internal database and that received/sent out by handlers - as per our discussion with jj and lewis - we introduced to the `ExternalScope` interface for clarity

src/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,111 @@
...
+import SDKStandardComponents from '@mojaloop/sdk-standard-components'
+import { ExternalScope } from '../../../lib/scopes'
+
+/*
+ * Interface for Consent Credential resource type
+ */
+export interface ConsentCredential {
+  credentialType: string;
+  credentialStatus: string;
+  credentialPayload?: string | undefined;
```
eoln, line 45:
```suggestion
  credentialPayload?: string | null;
```

src/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,111 @@
...
+export async function updateConsentCredential (
+  consent: Consent,
+  credential: ConsentCredential): Promise<Consent> {
+  // Update consent credentials
+  consent.credentialType = credential.credentialType
+  consent.credentialStatus = credential.credentialStatus
+  consent.credentialChallenge = credential.credentialChallenge
+
+  // Update in database
+  await consentDB.update(consent)
```
eoln, line 76: does it update only `credential*` related fields or all consent fields?

eoln: maybe we should implement a guard here to be sure which fields of consent are updated

src/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,111 @@
...
+    requestId: consent.id,
+    initiatorId: consent.initiatorId as string,
+    participantId: consent.participantId as string,
+    scopes,
+    credential: {
+      id: null,
+      credentialType: consent.credentialType as 'FIDO',
+      status: consent.credentialStatus as 'PENDING',
+      challenge: {
+        // TODO: Ask Lewis about array buffer
```
eoln, line 101: there is no ArrayBuffer anymore - use string

src/domain/consents/{ID}/generateChallenge.ts
akapur99: I believe #21 by @ramanmangla  deals with this

akapur99:
The failing checks are due to a dependency in `sdk-standard-components` - I've submitted a PR there to fix those

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Request, ResponseToolkit, ResponseObject } from '@hapi/hapi'
+import { consentDB, scopeDB } from '../../../../lib/db'
+import { Consent } from '../../../../model/consent'
+import Logger from '@mojaloop/central-services-logger'
+import { convertScopesToExternal } from '../../../../lib/scopes'
+import { Scope } from '../../../../model/scope'
+
+// Asynchronously deals with generating challenge, updating consent db
+//  and making outgoing PUT consent/{ID} call
+export async function postBackground (
```
jgeewax, line 42: Where did the naming convention on these come from? Is it a hapi standard or are we choosing them? They're way less descriptive than your other method names which are much clearer.

test/unit/domain/consents/{ID}/generateChallenge.test.ts
```diff
@@ -0,0 +1,214 @@
...
+const mockLoggerPush = jest.spyOn(Logger, 'push')
+const mockLoggerError = jest.spyOn(Logger, 'error')
+
+/*
+ * Mock Request Resources
+ */
+// @ts-ignore
+const request: Request = {
+  headers: {
+    fspiopsource: 'pisp-2342-2233',
```
jgeewax, line 57: Do we need to keep these all lower-case? Or can we do proper capitalization (e.g., `FSPIOPSource`) ?

akapur99: so in a different repo I found this being accessed as lowercase - but this is also just a test so either works

src/server/handlers/consents/{ID}/generateChallenge.ts
akapur99: So @eoln  and I were working on this together - this is what used to be within post() within a setImmediate call - not completely sure if this needs to be more descriptive as it is essentially just the 'background/async' section of the post handler?

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,90 @@
...
+ * contain the `{ID}` that was used in the `POST /consents`.
+ * Called by a `PISP`to request a challenge from the `auth-service`, which
+ * will be returned to the PISP via `PUT /consents/{ID}`
+ */
+export async function post (
+  request: Request, h: ResponseToolkit): Promise<ResponseObject> {
+  const id = request.params.id
+
+  // Fetch consent using ID
+  let consent: Consent = null as unknown as Consent
```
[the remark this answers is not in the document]
akapur99: makes sense - I'll change it

src/server/handlers/consents/{ID}/generateChallenge.ts
eoln: you are right @jgeewax, the name is ugly

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,114 @@
...
+import { Request, ResponseToolkit, ResponseObject } from '@hapi/hapi'
+import { consentDB, scopeDB } from '../../../../lib/db'
+import { Consent } from '../../../../model/consent'
+import Logger from '@mojaloop/central-services-logger'
+import { convertScopesToExternal } from '../../../../lib/scopes'
+import { Scope } from '../../../../model/scope'
+
+// Asynchronously deals with generating challenge, updating consent db
+//  and making outgoing PUT consent/{ID} call
+export async function postBackground (
```
eoln, line 42:
```suggestion
export async function generateChallengeAndPutConsentId (
```

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  // If consent is invalid, return 400 code
+  if (!isConsentRequestInitiatedByValidSource(request, consent)) {
+    return h.response().code(400)
+  }
+
+  // Asynchronously deals with generating challenge, updating consent db
+  //  and making outgoing PUT consent/{ID} call
+  postBackground(request, consent, id)
```
eoln, line 109:
```suggestion
  generateChallengeAndPutConsentId(request, consent, id)
```

test/unit/domain/consents/{ID}/generateChallenge.test.ts
akapur99: Actually @jgeewax since we should be following the enum declaration the correct usage is 'fispiop-source' I think 
Have made a change to reflect that

src/lib/requests.ts
```diff
@@ -30,17 +30,20 @@
  --------------
  ******/
 /* istanbul ignore file */
-// Testing will be covered in PR #11
+// Testing will be covered in Ticket #354
 
 import { config } from './config'
+import { ThirdpartyRequests, BaseRequestConfigType } from '@mojaloop/sdk-standard-components'
+import Logger from '@mojaloop/central-services-logger'
 // @ts-ignore
```
ramanmangla, line 38: Remove `@ts-ignore` and extra comment in line 39.

akapur99: addressed

src/lib/requests.ts
```diff
@@ -30,17 +30,19 @@
  --------------
  ******/
 /* istanbul ignore file */
-// Testing will be covered in PR #11
+// Testing will be covered in Ticket #354
```
jgeewax, line 33:
```suggestion
// Testing will be covered in #354
```

src/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+export async function putConsentId (
+  consent: Consent,
+  request: Request,
+  scopes: ExternalScope[]):
+  Promise<SDKStandardComponents.GenericRequestResponse | undefined> {
+  // Construct body of outgoing request
+  const body = {
+    requestId: consent.id,
+    initiatorId: consent.initiatorId as string,
```
eoln, line 101: why do we need this typecasting?

eoln: according to Consent interface definition, it should be a string
https://github.com/akapur99/auth-service/blob/consent1/src/model/consent/consent.ts#L51
same for other castings below.

src/domain/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,118 @@
+/* istanbul ignore file */
```
eoln, line 1: we don't need {ID} in domain's  folder structure -> it is only required for handlers

src/lib/challenge.ts
```diff
@@ -2,9 +2,14 @@
  License
  --------------
  Copyright © 2020 Mojaloop Foundation
```
eoln, line 4: could you update the rest of the files with this compact License text?

src/lib/requests.ts
```diff
@@ -30,17 +30,19 @@
...
 import { config } from './config'
-// @ts-ignore
-import { Logger, ThirdpartyRequests, BaseRequestConfigType } from '@mojaloop/sdk-standard-components'
+import { ThirdpartyRequests, BaseRequestConfigType } from '@mojaloop/sdk-standard-components'
+import Logger from '@mojaloop/central-services-logger'
+// TODO: Switch to sdk-standard-components Logger once implemented
 
 // Config file to instantiate ThirdPartyRequest object
 const configRequest: BaseRequestConfigType = {
   dfspId: config.get('PARTICIPANT_ID') as string,
```
eoln, line 42: why typecasting, if you would like to be sure you have a string in execution time use `.toString()` -> typecasting is only compilation time construct

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,130 @@
...
+      const credential: ConsentCredential = {
+        credentialChallenge: challengeValue,
+        credentialStatus: 'PENDING',
+        credentialType: 'FIDO'
+      }
+
+      consent = await updateConsentCredential(consent, credential)
+    } else if (consent.credentialStatus === 'ACTIVE') {
+      // TODO: Error handling here - dealt with in #355
+      Logger.error('ACTIVE credential consent has requested challenge')
```
eoln, line 76: we should throw here NotImplementedYetError, elsewhere there is a risk we will not implement it

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,130 @@
...
+    const scopesRetrieved: Scope[] = await scopeDB.retrieveAll(id)
+    const scopes = convertScopesToExternal(scopesRetrieved)
+
+    // Outgoing call to PUT consents/{ID}
+    await putConsentId(consent, request, scopes)
+  } catch (error) {
+    Logger.push(error)
+    // eslint-disable-next-line max-len
+    Logger.error(`Error: Outgoing call with challenge credential NOT made to  PUT consent/${id}`)
+    // TODO: Decide on error handling HERE -  dealt with in future ticket #355
```
eoln, line 90: Let rethrow the received error instead of keeping it silent. In the next story we will transfer it to PUT .../error

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,130 @@
...
+ * credential for the given Consent object. The `{ID}` in the URI should
+ * contain the `{ID}` that was used in the `POST /consents`.
+ * Called by a `PISP`to request a challenge from the `auth-service`, which
+ * will be returned to the PISP via `PUT /consents/{ID}`
+ */
+export async function post (
+  request: Request, h: ResponseToolkit): Promise<ResponseObject> {
+  const id = request.params.id
+
+  // Fetch consent using ID
```
eoln, line 104: according to the discussion we have had yesterday during a standup with @jgeewax and @ramanmangla let move DB related validation to background function: generateChallengeAndPutConsentId and return 202 asap

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,130 @@
...
+    return h.response().code(400)
+  }
+
+  // Asynchronously deals with generating challenge, updating consent db
+  //  and making outgoing PUT consent/{ID} call
+  generateChallengeAndPutConsentId(request, consent, id)
+  // intentionally not await-ing we want to run it in background
+
+  // Return Success code informing source: request received
+  return h.response().code(202)
```
eoln, line 129:
```suggestion
  return h.response().code(Enum.Http.ReturnCodes.ACCEPTED.CODE)
```

and add at the begin of file:
`import { Enum } from '@mojaloop/central-services-shared'`
Use always status error codes from this Enum

src/server/handlers/consents/{ID}/generateChallenge.ts
akapur99: @eoln if we throw an error here or in the previous comment - there will be an unhandled promise rejection issue as nothing is handling the POST handler's errors

src/server/handlers/consents/{ID}/generateChallenge.ts
akapur99: same comment as below - re: handling of errors @eoln

src/server/handlers/consents/{ID}/generateChallenge.ts
akapur99: but we want to return a 400 code if request is bad - so that would need to be done synchronously, correct? @ramanmangla  @eoln  @jgeewax

src/domain/consents/{ID}/generateChallenge.ts
akapur99: I'm confused so then should the structure just be 
domain/consents/generateChallenge?

src/lib/requests.ts
akapur99: was typecasting to avoid a TS error
Even if i use .toString - same error is thrown - as TS cannot parse the type 
also @spikerheado1234  in your lib config file you've done the same - did you mean to do that? perhaps you should also be doing .toString because it seems if you don't you're getting returned an object?

src/lib/challenge.ts
akapur99: done

src/domain/consents/{ID}/generateChallenge.ts
akapur99: most of the places where there is typecasting is just for TS benefit

akapur99: this is because for the fields being typecast here
their type is `string | undefined` 
so typescript is not ok with assigning them to a field with type `string` as there is the possibility of them being undefined 
however since the handler checks all this *we know* that they must be `string`

src/server/handlers/consents/{ID}/generateChallenge.ts
eoln: Yes, you are right, normally we shouldn't allow promise rejections to bubble up, but unhandled promise rejection will be temporary - until we got PUT .../error implemented. IMHO it is better to have comment TODO in code, and also unhandled promise rejection/exception which will help us remember that we have some unfinished flow

src/server/handlers/consents/{ID}/generateChallenge.ts
eoln: if validation is expensive -> if there is a need to do any SQL based query -> we should return 202 ASAP and do rest in background function. If any validation fails we should report about via PUT/.../error . I got the info from @elnyry-sam-k that we can use 6xxxx error codes for PISP: 6x00 - 6x39 dedicated to the API and then rest to the Schemes -> take a look for existing ones: https://github.com/mojaloop/mojaloop-specification/blob/master/documents/Logical%20Data%20Model.md#46-error-codes

src/lib/requests.ts
eoln: I am starting to love this: `dfspId: config.get('PARTICIPANT_ID').toString() as string`
maybe using https://www.npmjs.com/package/@types/convict would solve the problem?

src/server/handlers/consents/{ID}/generateChallenge.ts
akapur99: makes sense have addressed this

src/server/handlers/consents/{ID}/generateChallenge.ts
akapur99: makes sense will fix this by end of day

src/lib/requests.ts
akapur99: i tried but it didn't
so finally have reverted back to initial implementation

src/server/handlers/consents/{ID}/generateChallenge.ts
akapur99: doen

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -0,0 +1,140 @@
...
+ * will be returned to the PISP via `PUT /consents/{ID}`
+ */
+export async function post (
+  request: Request, h: ResponseToolkit): Promise<ResponseObject> {
+  const id = request.params.id
+
+  // Asynchronously deals with validating request,
+  //  generating challenge, updating consent db
+  //  and making outgoing PUT consent/{ID} call
+  generateChallengeAndPutConsentId(request, id)
```
jgeewax, line 135:
```suggestion
  generateChallengeAndPutConsent(request, id)
```

Should we use a slightly different name? Or are we putting an ID?

akapur99: ah we're making a call to PUT consent/{ID} 
but either works?

akapur99: done

PR #12, opened 2020-07-03
Feature/341: Add DB models and unit tests for Scope resource

ramanmangla:
In reference to ticket number #341. Addresses tasks 3 and 4.

This integrates the existing database migrations setup with the project by building data querying models on top of it.
 - Implements querying models for creating, retrieving and deleting Scope resources from the DB
 - Adds unit tests using in memory SQLite to verify each model's functionality

Related future work will be addressed in following tickets:
 - `#355` Consistent Error Handling in Auth Service

Main Work Files:
 - `src/model/scope/scope.ts`
 - `src/model/scope/index.ts`
 - `test/unit/model/scope.test.ts`
 - `src/lib/db.ts`


migrations/migration.template
```diff
@@ -0,0 +1,9 @@
+import * as knex from 'knex'
+
+export function up (knex: Knex): Knex.SchemaBuilder {
+
```
eoln, line 4: maybe we can put the invocation of createTableifNotExist here with empty callback function with some additional comments on how the `up` should be implemented?

migrations/migration.template
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export function down (knex: Knex): Knex.SchemaBuilder {
+
```
eoln, line 8: same as above but with the usage of `dropTableifExist`

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      // Action
+      await consentDB.register(partialConsent)
+
+      // Assertion
+      const consents: Consent[] = await Db<Consent>('Consent')
+        .select('*')
+        .where({
+          id: partialConsent.id
+        })
+
```
eoln, line 83: if we are expecting `createdAt` to be present in the query result, we should assert it before delete, and add a comment why we  want to delete this property

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      delete consents[0].createdAt
```
eoln, line 109: same as above `createdAt`

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      await consentDB.updateCredentials(completeConsent)
+
+      // Assertion
+      const consents: Consent[] = await Db<Consent>('Consent')
+        .select('*')
+        .where({
+          id: completeConsent.id
+        })
+
+      delete consents[0].createdAt
```
eoln, line 151: same as above `createdAt`

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    beforeEach(async (): Promise<void> => {
+      await Db<Consent>('Consent').del()
+      await Db<Consent>('Consent')
+        .insert(completeConsent)
+    })
+
+    it('retrieves an existing consent', async (): Promise<void> => {
+      // Action
+      const consents: Consent[] = await consentDB.retrieve(completeConsent.id)
+
```
eoln, line 168: same as above `createdAt`

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      await scopeDB.register(tempScope)
+
+      // Assertion
+      const scopes: Scope[] = await Db<Scope>('Scope')
+        .select('*')
+        .where({
+          consentId: partialConsent.id
+        })
+
+      delete scopes[0].id
```
eoln, line 95: same as above `createdAt`

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      // Setup
+      await Db<Consent>('Consent')
+        .insert(partialConsent)
+      await Db<Scope>('Scope')
+        .insert(tempScope)
+
+      // Action
+      const scopes: Scope[] = await scopeDB.retrieve(tempScope.consentId)
+
+      delete scopes[0].id
```
eoln, line 125: same as above `createdAt`

package.json
```diff
@@ -7,6 +7,9 @@
   "files": [
     "dist"
   ],
+  "config": {
+    "knex": "--knexfile ./config/knexfile.ts"
+  },
   "scripts": {
     "audit:resolve": "SHELL=sh resolve-audit",
```
stevenwjy, line 14: Btw, just happened to see this, but you can add `--production` to both `audit:resolve` and `audit:check` to avoid having ~42k lines of `audit-resolve.json`.

ramanmangla: Thanks for the suggestion @stevenwjy. @lewisdaly @eoln Any thoughts on this?

migrations/migration.template
ramanmangla: @eoln should @spikerheado1234 handle this? He worked on the migrations.

migrations/migration.template
ramanmangla: @eoln @spikerheado1234 same as above

migrations/migration.template
eoln: it would be nice @spikerheado1234 if you could help @ramanmangla :-) 
Sorry for the late request. I haven't noticed it on time, and your PR has been already merged.

src/model/scope/scope.ts
```diff
@@ -0,0 +1,72 @@
...
+  // Knex instance
+  private Db: Knex
+
+  public constructor (dbInstance: Knex) {
+    this.Db = dbInstance
+  }
+
+  // Add Scope resource
+  public register (scope: Scope): Promise<Scope[]> {
+    return this.Db<Scope>('Scope')
```
jgeewax, line 55: Same feedback for the other stuff here (re errors).

src/model/scope/scope.ts
```diff
@@ -0,0 +1,72 @@
...
+export class ScopeDB {
+  // Knex instance
+  private Db: Knex
+
+  public constructor (dbInstance: Knex) {
+    this.Db = dbInstance
+  }
+
+  // Add Scope resource
+  public register (scope: Scope): Promise<Scope[]> {
```
jgeewax, line 54:
```suggestion
  public register (scope: Scope): Promise<Scope> {
```

src/model/scope/scope.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.Db<Scope>('Scope')
+      .insert(scope)
+  }
+
+  // Retrieve Scope by Consent ID
+  public retrieve (consentId: string): Promise<Scope[]> {
```
jgeewax, line 60:
```suggestion
  public retrieve (consentId: string): Promise<Scope> {
```

src/model/scope/scope.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.Db<Scope>('Scope')
+      .select('*')
+      .where({ consentId: consentId })
```
jgeewax, line 63: Use NotFoundError as well here

src/model/scope/scope.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  // Delete Scope by Consent ID
+  public delete (consentId: string): Promise<Scope[]> {
```
jgeewax, line 67:
```suggestion
  public delete (consentId: string): Promise<Scope> {
```

src/model/scope/scope.ts
```diff
@@ -28,6 +28,7 @@
  ******/
 
 import Knex from 'knex'
+import NotFoundError from '../error'
```
eoln, line 38: let use `errors` instead `error`

src/model/scope/scope.ts
```diff
@@ -51,20 +52,37 @@ export class ScopeDB {
...
+  public async register (scope: Scope): Promise<number> {
+    let insertCount: number[]
+
+    try {
+      // Returns array containing number of inserted rows
+      insertCount = await this
+        .Db<Scope>('Scope')
+        .insert(scope)
+    } catch (err) {
+      throw Error(err)
```
eoln, line 71: why error reformating? Is knex returning non-standard errors?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  public constructor (dbInstance: Knex) {
+    this.Db = dbInstance
+  }
+
+  // Add initial Consent parameters
+  public async register (consent: Consent): Promise<number> {
+    let insertCount: number[]
+
+    try {
```
jgeewax, line 85: Isn't this try block effectively doing nothing? It catches an error and then rethrows it, right?

src/model/consent/consent.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    return consents
+  }
+
+  // Delete Consent by ID
+  // Deleting Consent automatically deletes associates scopes
+  public delete (id: string): Promise<number> {
+    // Returns number of deleted rows
+    return this.Db<Consent>('Consent')
```
jgeewax, line 143: Shouldn't we throw a NotFound error if the result is 0 items deleted?

src/model/scope/scope.ts
```diff
@@ -0,0 +1,91 @@
...
+  public async register (scope: Scope): Promise<number> {
+    // Returns array containing number of inserted rows
+    const insertCount: number[] = await this
+      .Db<Scope>('Scope')
+      .insert(scope)
+
+    return insertCount[0]
+  }
+
+  // Retrieve Scope by Consent ID
```
jgeewax, line 71:
```suggestion
  // Retrieve Scopes by Consent ID
```

src/model/scope/scope.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    if (scopes.length === 0) {
+      throw new NotFoundError('Scope', consentId)
+    }
+
+    return scopes
+  }
+
+  // Delete Scope by Consent ID
+  public delete (consentId: string): Promise<number> {
```
jgeewax, line 86: This deletes **all** scopes for a given consent, right? What if you just want to remove one of them?

src/model/consent/consent.ts
ramanmangla: @jgeewax Sorry for the confusion but this file corresponds to the other PR. It was changed earlier in the PR.

src/model/scope/scope.ts
ramanmangla: Based on the sequence diagrams and the immutability you mentioned about the DB, my understanding was that one couldn't delete a specific scope. We may need to have some kind of Scope Identifier for deleting a specific one.

ramanmangla: Adding on to that, we automatically delete the associated Scopes through cascade on deleting a Consent.

ramanmangla: If we add a standard Scope identifier, we should also add retrieval of specific Scopes.

jgeewax: It's fine that it deletes all of them, but it should be `deleteAll` or something like that -- otherwise the name of the method doesn't quite line up with the behavior.

src/model/consent/consent.ts
jgeewax: OK - can we get ... a PR that reflects the changes that we're talking about here then? Any delta in here should be considered "fair game" for comments :-(

ramanmangla: Yes, it's all merged across the board. I hope it helps. I'll ensure to put a PR against a feature branch if some work builds off of that feature branch in the future. Maybe we could get the Consent stuff merged in first for proper deltas?
Apologies again :(

test/unit/model/scope.test.ts
```diff
@@ -130,13 +130,13 @@ describe('src/model/scope', (): void => {
...
-      await expect(scopeDB.retrieve(partialConsent.id))
+      await expect(scopeDB.retrieveAll(partialConsent.id))
         .rejects.toThrowError('NotFoundError: Scope for ConsentId 1234')
     })
   })
 
-  describe('delete', (): void => {
-    it('deletes existing scope from the database', async (): Promise<void> => {
+  describe('deleteAll', (): void => {
+    it('deletes existing scopes from the database', async (): Promise<void> => {
```
eoln, line 139:
```suggestion
    it('deletes only existing scopes from the database', async (): Promise<void> => {
```

test/unit/model/scope.test.ts
```diff
@@ -154,7 +154,7 @@ describe('src/model/scope', (): void => {
       expect(scopes.length).toEqual(2)
 
       // Action
-      await scopeDB.delete(tempScope.consentId)
+      await scopeDB.deleteAll(tempScope.consentId)
```
eoln, line 158: after `deleteAll` you should check that in the database there are no scopes for `tempScope.consentId`

test/unit/model/scope.test.ts
```diff
@@ -102,13 +102,32 @@ describe('src/model/scope', (): void => {
       expect(scopes[0]).toEqual(expect.objectContaining(tempScope))
     })
 
+    it('adds multiple scopes for an existing consent to the database', async (): Promise<void> => {
```
eoln, line 105: in the test above `adds scope for an existing consent to the database` you should test that only one scope has been added

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,198 @@
...
+          consentId: partialConsent.id
+        })
+
+      expect(scopes.length).toEqual(2)
+      expect(scopes[0].consentId === scopes[1].consentId).toEqual(true)
+      expect(scopes[0].action !== scopes[1].action).toEqual(true)
+    })
+
+    it('returns an error on adding a scope for a non-existent consent', async (): Promise<void> => {
+      await expect(scopeDB.register(tempScope)).rejects.toThrowError()
```
jgeewax, line 125: Check for the specific error class here?

ramanmangla: This is a constraint violation. Hence, the error is DB dependent and not defined by the code.
We don't have an error class for this.

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,267 @@
...
+  // Reset table for new test
+  beforeEach(async (): Promise<void> => {
+    await Db<Consent>('Consent').del()
+  })
+
+  describe('register', (): void => {
+    it('adds consent with partial info to the database', async (): Promise<void> => {
+      // Action
+      const numInserted: number = await consentDB.register(partialConsent)
```
eoln, line 89:
```suggestion
      const insertCount: number = await consentDB.register(partialConsent)
```

test/unit/model/consent.test.ts
```diff
@@ -0,0 +1,267 @@
...
+
+      expect(numInserted).toEqual(1)
```
eoln, line 91:
```suggestion
      expect(insertCount).toEqual(1)
```

src/model/scope/scope.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // Retrieve Scopes by Consent ID
+  public async retrieveAll (consentId: string): Promise<Scope[]> {
+    const scopes: Scope[] = await this
+      .Db<Scope>('Scope')
+      .select('*')
+      .where({ consentId: consentId })
+
+    if (scopes.length === 0) {
```
jgeewax, line 79:
```suggestion
    // We don't distinguish between a Consent that exists with 0 scopes and a Consent that does not exist.
    if (scopes.length === 0) {
```

src/model/scope/scope.ts
```diff
@@ -0,0 +1,99 @@
...
+      throw new NotFoundError('Scope', consentId)
```
jgeewax, line 80:
```suggestion
      throw new NotFoundError('Consent's Scopes', consentId)
```

src/model/scope/scope.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  // Delete Scopes by Consent ID
+  public async deleteAll (consentId: string): Promise<number> {
+    const deleteCount: number = await this
+      .Db<Scope>('Scope')
+      .where({ consentId: consentId })
+      .del()
+
+    if (deleteCount === 0) {
+      throw new NotFoundError('Scope', consentId)
```
jgeewax, line 94: Same error message as above.

src/model/scope/scope.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // Delete Scopes by Consent ID
+  public async deleteAll (consentId: string): Promise<number> {
+    const deleteCount: number = await this
+      .Db<Scope>('Scope')
+      .where({ consentId: consentId })
+      .del()
+
+    if (deleteCount === 0) {
```
jgeewax, line 93: Same comment as above that we don't distinguish between a missing Consent and a Consent with no Scopes.

src/model/scope/scope.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    return scopes
+  }
+
+  // Delete Scopes by Consent ID
+  public async deleteAll (consentId: string): Promise<number> {
+    const deleteCount: number = await this
+      .Db<Scope>('Scope')
+      .where({ consentId: consentId })
```
jgeewax, line 90:
```suggestion
      .where({ consentId })
```

src/model/scope/scope.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    return insertCount[0]
+  }
+
+  // Retrieve Scopes by Consent ID
+  public async retrieveAll (consentId: string): Promise<Scope[]> {
+    const scopes: Scope[] = await this
+      .Db<Scope>('Scope')
+      .select('*')
+      .where({ consentId: consentId })
```
jgeewax, line 77:
```suggestion
      .where({ consentId })
```

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,198 @@
...
+        .where({
+          consentId: partialConsent.id
+        })
+
+      expect(scopes.length).toEqual(1)
+      expect(scopes[0].id).toEqual(expect.any(Number))
+      expect(scopes[0]).toEqual(expect.objectContaining(tempScope))
+    })
+
+    it('adds multiple scopes for an existing consent to the database', async (): Promise<void> => {
```
jgeewax, line 105: Let's add a case of adding an empty array of scopes.

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    })
+
+    it('throws an error on retrieving of non-existent scopes', async (): Promise<void> => {
+      // Setup
+      await Db<Consent>('Consent')
+        .insert(partialConsent)
+
+      // Action
+      await expect(scopeDB.retrieveAll(partialConsent.id))
+        .rejects.toThrowError('NotFoundError: Scope for ConsentId 1234')
```
jgeewax, line 153: We should be able to test just for "NotFoundError" (the type) rather than the exact error message. Error messages tend to change over time, so we don't want to anchor on those.

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    })
+
+    it('throws an error on deleting non-existent scopes', async (): Promise<void> => {
+      // Setup
+      await Db<Consent>('Consent')
+        .insert(partialConsent)
+
+      // Action
+      await expect(scopeDB.deleteAll(partialConsent.id))
+        .rejects.toThrowError('NotFoundError: Scope for ConsentId 1234')
```
jgeewax, line 195: Same as above.

src/model/scope/scope.ts
```diff
@@ -0,0 +1,107 @@
...
+  // Knex instance
+  private Db: Knex
+
+  public constructor (dbInstance: Knex) {
+    this.Db = dbInstance
+  }
+
+  // Add a single Scope or an array of Scopes
+  public async insert (scopes: Scope | Scope[]): Promise<boolean> {
+    if (Array.isArray(scopes) && scopes.length === 0) {
```
eoln, line 64: this is input validation phase, so I would rename the error exception to `InvalidArgumentError`

src/model/scope/scope.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (scopes.length === 0) {
+      throw new NotFoundError('Consent\'s Scopes', consentId)
+    }
+
+    return scopes
+  }
+
+  // Delete Scopes by Consent ID
+  public async deleteAll (consentId: string): Promise<number> {
+    const deleteCount: number = await this
```
eoln, line 94: what if `consentId` is empty string?

eoln: if we decided to do input validation let be consistent

ramanmangla: There is no input validation in the models @eoln . The only place there is validation is for Empty array in insert because it returns a "SQL MISUSE: not an error" exception which is also dependent on the DB. So I added it to standardize it.

jgeewax: If you pass in garbage, I'm not sure if it's the model's job to deal with your garbage. The Handler layer should look at this and decide to reject bad input. If the result of an empty string was a bunch of "technically correct" but not quite correct data returned, then I'd say we should defend against that. But for an empty string you'll just get back zero results, so that should be fine.

jgeewax, line 94: Is this the same for MySQL and SQLite ?

ramanmangla: Yep, it's the same.

src/model/scope/scope.ts
ramanmangla: I removed the error @eoln and just defined a standard return type for empty arrays here to avoid any inconsistencies among the DBs.

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  id: '948',
+  initiatorId: 'pisp-2342-2013',
+  participantId: 'dfsp-3333-2773'
+}
+
+/*
+ * Mock Scope Resources
+ */
+const tempScope1: Scope = {
+  consentId: '1234',
```
eoln, line 55:
```suggestion
  consentId:  partialConsent1.id,
```

test/unit/model/scope.test.ts
```diff
@@ -0,0 +1,191 @@
...
+
+      // Assertion
+      const scopes: Scope[] = await Db<Scope>('Scope')
+        .select('*')
+        .where({
+          consentId: partialConsent1.id
+        })
+
+      expect(scopes.length).toEqual(1)
+      expect(scopes[0].id).toEqual(expect.any(Number))
```
eoln, line 115:
```suggestion
      expect(scopes[0]).toEqual(tempScope1)
```

PR #14, opened 2020-07-09
Separate module file for all challenge generation/validation

akapur99:
Small PR just to prevent blockage on further tickets such as #340
Spinoff of PR #11 dealing with ticket #339 : Handles generation of challenge string

Related future work will be addressed in following ticket:
#354 End-to-end BDD Testing for Auth-Service Endpoints

### DIFF
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -44,7 +44,7 @@ export async function generate (size: number = 32): Promise<string> {
     const buf = await randomBytesAsync(Math.round(Math.abs(size)))
     return buf.toString('base64')
   } catch (error) {
-    Logger.error('Unable to generate challenge string')
+    Logger.push(error).error('Unable to generate challenge string')
     throw error
   }
 }
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -1,0 +1,46 @@
+/*****
+ License
+ --------------
+ Copyright © 2020 Mojaloop Foundation
+ The Mojaloop files are made available by the Mojaloop Foundation under the Apache License, Version 2.0 (the 'License') and you may not use these files except in compliance with the License. You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


spikerheado1234:
Lgtm!

src/lib/challenge.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Helper function which uses the crypto library to generate
+ * a secure random challenge string (Base 64 encoding) of given size
+ * @param size Integer value of how many bytes should generated, 32 by default
+ */
+export async function generate (size: number = 32): Promise<string> {
+  try {
+    const buf = await randomBytesAsync(Math.round(Math.abs(size)))
+    return buf.toString('base64')
+  } catch (error) {
+    Logger.error('Unable to generate challenge string')
```
eoln, line 47:
```suggestion
    Logger.push(error).error('Unable to generate challenge string')
```

test/unit/lib/challenge.test.ts
```diff
@@ -0,0 +1,13 @@
+import { generate } from '../../../src/lib/challenge'
+
+describe('Challenge Generation', (): void => {
+  it('Should return a 32 byte string', async (): Promise<void> => {
```
eoln, line 4:
```suggestion
  it('Should return a 32 byte string by default', async (): Promise<void> => {
```

test/unit/lib/challenge.test.ts
```diff
@@ -0,0 +1,13 @@
...
+    const challenge = await generate()
+    expect(Buffer.byteLength(challenge, 'base64')).toBe(32)
+  })
+
+  it('Should return a 64 byte string', async (): Promise<void> => {
```
eoln, line 9:
```suggestion
  it('Should return a 64 byte string if argument passed', async (): Promise<void> => {
```

test/unit/lib/challenge.test.ts
```diff
@@ -0,0 +1,13 @@
+import { generate } from '../../../src/lib/challenge'
```
eoln, line 1: there should be cases when `size` parameter is nonpositive number

PR #16, opened 2020-07-13
Feature/342: Add Signature Verification function to Challenge module

ramanmangla:
In reference to ticket number #342.

This is a small PR that adds a function to verify incoming digital signatures.
 - Allows RSA and Elliptic keys signature verification
 - Uses SHA256 for hashing the signature
 - Tested on RSA 2048 and EC secp256k1 keys

Related future work will be addressed in following ticket:

Main Work Files:
 - `src/lib/challenge.ts`
 - `test/unit/lib/challenge.test.ts`

### DIFF
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -32,7 +32,7 @@
  */
 
 import util from 'util'
-import crypto, { KeyObject } from 'crypto'
+import crypto from 'crypto'
 import { Logger } from '@mojaloop/central-services-logger'
 
 // Async promisified randomBytes function
@@ -53,16 +53,16 @@ export async function generate (size: number = 32): Promise<string> {
   }
 }
 
-/*
+/**
  * Helper function to validate signatures using public key
- * @param challenge Base64 challenge string
+ * @param challenge UTF-8 challenge string
  * @param signature Base64 sign string
  * @param publicKey Base64 RSA/EC:secp256k1 key string or KeyObject for verification
  */
-export async function verifySign (
+export function verifySign (
   challenge: string,
   signature: string,
-  publicKey: string | KeyObject): Promise<boolean> {
+  publicKey: string | crypto.KeyObject): boolean {
   // Hash using SHA256
   const verifier: crypto.Verify = crypto.createVerify('SHA256')
 
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -57,13 +57,13 @@ export async function generate (size: number = 32): Promise<string> {
 /**
  * Helper function to validate signatures using public key
  * @param challenge UTF-8 challenge string
- * @param signature Base64 sign string
+ * @param signature Base64 signature string
  * @param publicKey PEM Base64 Public key string or KeyObject for verification
  *
  * Currently, the implementation focuses on RSA 2048 and ECDSA:secp256k1 keys.
  * Support for additional keys can be extended further.
  */
-export function verifySign (
+export function verifySignature (
   challenge: string,
   signature: string,
   publicKey: string | crypto.KeyObject): boolean {
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -58,21 +58,24 @@ export async function generate (size: number = 32): Promise<string> {
  * Helper function to validate signatures using public key
  * @param challenge UTF-8 challenge string
  * @param signature Base64 sign string
- * @param publicKey Base64 RSA/EC:secp256k1 key string or KeyObject for verification
+ * @param publicKey PEM Base64 Public key string or KeyObject for verification
+ *
+ * Currently, the implementation focuses on RSA and ECDSA:secp256k1 keys.
+ * Support for additional keys can be extended further.
  */
 export function verifySign (
   challenge: string,
   signature: string,
   publicKey: string | crypto.KeyObject): boolean {
   try {
-    // Hash using SHA256
+    // Digest Algorithm
     const verifier: crypto.Verify = crypto.createVerify('SHA256')
-
+    // Hashing the challenge string
     verifier.update(challenge)
 
     return verifier.verify(publicKey, signature, 'base64')
   } catch (error) {
-    Logger.push(error).error('Unable to verify signature')
+    Logger.error('Unable to verify signature')
     throw error
   }
 }
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -70,3 +70,5 @@ export function verifySign (
 
   return verifier.verify(publicKey, signature, 'base64')
 }
+
+// Can createVerify be declared once in a beforeEach or global?
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -63,12 +63,15 @@ export function verifySign (
   challenge: string,
   signature: string,
   publicKey: string | crypto.KeyObject): boolean {
-  // Hash using SHA256
-  const verifier: crypto.Verify = crypto.createVerify('SHA256')
+  try {
+    // Hash using SHA256
+    const verifier: crypto.Verify = crypto.createVerify('SHA256')
 
-  verifier.update(challenge)
+    verifier.update(challenge)
 
-  return verifier.verify(publicKey, signature, 'base64')
+    return verifier.verify(publicKey, signature, 'base64')
+  } catch (error) {
+    Logger.push(error).error('Unable to verify signature')
+    throw error
+  }
 }
-
-// Can createVerify be declared once in a beforeEach or global?
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -19,13 +19,17 @@
  - Name Surname <name.surname@gatesfoundation.com>
 
  - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+ - Raman Mangla <ramanmangla@google.com>
 
  --------------
  ******/
 
 /* istanbul ignore file */
 
-// Flag to igore BDD testing, which will be dealt with in a later ticket
+/**
+ * Flag to ignore BDD testing, which will be dealt
+ * with in a later ticket
+ */
 
 import util from 'util'
 import crypto from 'crypto'
@@ -47,4 +51,23 @@ export async function generate (size: number = 32): Promise<string> {
     Logger.push(error).error('Unable to generate challenge string')
     throw error
   }
+}
+
+/*
+ * Helper function to validate signatures using public key
+ */
+export async function verify (challenge: string, signature: string, key: string): Promise<boolean> {
+  // const challengeBuffer = Buffer.from(challenge, 'base64')
+  // const signatureBuffer = Buffer.from(signature, 'base64')
+
+  // return crypto.verify(signAlgorithm, challengeBuffer, key, signatureBuffer)
+  // const verify = crypto.createVerify('RSA-SHA256')
+  // verify.update(challenge, 'utf8')
+  // return verify.verify(key, signature)
+
+  const verifier: crypto.Verify = crypto.createVerify('sha256')
+
+  verifier.update(challenge)
+
+  return verifier.verify(key, signature, 'base64')
 }
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -53,21 +53,16 @@ export async function generate (size: number = 32): Promise<string> {
   }
 }
 
-/*
+/**
  * Helper function to validate signatures using public key
+ * @param challenge Base64 challenge string
+ * @param signature Base64 sign string
+ * @param publicKey Base64/PEM Public key string for verification
  */
-export async function verifySign (challenge: string, signature: string, key: string): Promise<boolean> {
-  // const challengeBuffer = Buffer.from(challenge, 'base64')
-  // const signatureBuffer = Buffer.from(signature, 'base64')
-
-  // return crypto.verify(signAlgorithm, challengeBuffer, key, signatureBuffer)
-  // const verify = crypto.createVerify('RSA-SHA256')
-  // verify.update(challenge, 'utf8')
-  // return verify.verify(key, signature)
-
-  const verifier: crypto.Verify = crypto.createVerify('sha256')
+export async function verifySign (challenge: string, signature: string, publicKey: string): Promise<boolean> {
+  const verifier: crypto.Verify = crypto.createVerify('RSA-SHA256')
 
   verifier.update(challenge)
 
-  return verifier.verify(key, signature, 'base64')
+  return verifier.verify(publicKey, signature, 'base64')
 }
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -27,8 +27,9 @@
 /* istanbul ignore file */
 
 /*
- * Flag to ignore BDD testing, which will be dealt
- * with in a later ticket
+ * This flag is to ignore BDD testing for model
+ * which will be addressed in the future in
+ * ticket #354
  */
 
 import util from 'util'
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -57,7 +57,7 @@ export async function generate (size: number = 32): Promise<string> {
  * Helper function to validate signatures using public key
  * @param challenge Base64 challenge string
  * @param signature Base64 sign string
- * @param publicKey Base64 RSA/EC:sec256k1 key string or KeyObject for verification
+ * @param publicKey Base64 RSA/EC:secp256k1 key string or KeyObject for verification
  */
 export async function verifySign (
   challenge: string,
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -57,9 +57,13 @@ export async function generate (size: number = 32): Promise<string> {
  * Helper function to validate signatures using public key
  * @param challenge Base64 challenge string
  * @param signature Base64 sign string
- * @param publicKey Base64/PEM Public key string or KeyObject for verification
+ * @param publicKey Base64 RSA/EC:sec256k1 key string or KeyObject for verification
  */
-export async function verifySign (challenge: string, signature: string, publicKey: string | KeyObject): Promise<boolean> {
+export async function verifySign (
+  challenge: string,
+  signature: string,
+  publicKey: string | KeyObject): Promise<boolean> {
+  // Hash using SHA256
   const verifier: crypto.Verify = crypto.createVerify('SHA256')
 
   verifier.update(challenge)
--- a/src/lib/challenge.ts
+++ b/src/lib/challenge.ts
@@ -60,7 +60,7 @@ export async function generate (size: number = 32): Promise<string> {
  * @param signature Base64 sign string
  * @param publicKey PEM Base64 Public key string or KeyObject for verification
  *
- * Currently, the implementation focuses on RSA and ECDSA:secp256k1 keys.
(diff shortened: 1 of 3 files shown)


src/lib/challenge.ts
```diff
@@ -48,3 +52,21 @@ export async function generate (size: number = 32): Promise<string> {
...
   }
 }
+
+/*
+ * Helper function to validate signatures using public key
+ * @param challenge Base64 challenge string
+ * @param signature Base64 sign string
+ * @param publicKey Base64 RSA/EC:sec256k1 key string or KeyObject for verification
+ */
+export async function verifySign (
```
lewisdaly, line 62: Is there any need for this to be async? I don't see any async functions being called here.

src/lib/challenge.ts
```diff
@@ -19,16 +19,20 @@
...
 
-// Flag to igore BDD testing, which will be dealt with in a later ticket
+/*
+ * Flag to ignore BDD testing, which will be dealt
+ * with in a later ticket
+ */
 
 import util from 'util'
-import crypto from 'crypto'
+import crypto, { KeyObject } from 'crypto'
```
akapur99, line 35: Hey why aren't you just using crypto.KeyObject to reference it? Instead of importing that separately?

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,130 @@ describe('Challenge Generation', (): void => {
     expect(Buffer.byteLength(challenge, 'base64')).toBe(8)
   })
 })
+
+describe('Signature Verification', (): void => {
+  // Repeat test with different keys
+  for (let index = 0; index < 4; index++) {
```
lewisdaly, line 33: I don't like wrapping `it` statements in loops like this. It makes it really hard to debug test failures. 

Perhaps internally you could generate a list of KeyPairs (`keyPairList`), and run a reducer function over these.

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,130 @@ describe('Challenge Generation', (): void => {
...
+      signer.update(challenge)
+
+      const sign = signer.sign(keyPair.privateKey, 'base64')
+
+      expect(await verifySign(challenge, sign, keyPair.publicKey)).toEqual(true)
+    })
+  }
+
+  // Repeat test with different keys
+  for (let index = 0; index < 4; index++) {
```
lewisdaly, line 59: Same as above

src/lib/challenge.ts
```diff
@@ -48,3 +52,21 @@ export async function generate (size: number = 32): Promise<string> {
...
+ * @param challenge Base64 challenge string
+ * @param signature Base64 sign string
+ * @param publicKey Base64 RSA/EC:sec256k1 key string or KeyObject for verification
+ */
+export async function verifySign (
+  challenge: string,
+  signature: string,
+  publicKey: string | KeyObject): Promise<boolean> {
+  // Hash using SHA256
+  const verifier: crypto.Verify = crypto.createVerify('SHA256')
```
akapur99, line 67: if any/all of these methods are `async` you should be using await + try/catch ?

akapur99: otherwise you'll have unresolved promises or unhandled errors passing through

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,130 @@ describe('Challenge Generation', (): void => {
...
+      })
+
+      const challenge = 'Crypto Auth service Yay!'
+      const signer = crypto.createSign('SHA256')
+
+      signer.update(challenge)
+
+      const sign = signer.sign(keyPair.privateKey, 'base64')
+
+      expect(await verifySign(challenge, sign, keyPair.publicKey)).toEqual(true)
```
lewisdaly, line 54: I tend to prefer assertions to not contain the actual test logic. It's much more readable to keep them separate.

```suggestion
      const sign = signer.sign(keyPair.privateKey, 'base64')
      const result = verifySign(challenge, sign, keyPair.publicKey)

      expect(result).toEqual(true)
```

Also this is `async`, but I don't think it needs to be (see above)

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,130 @@ describe('Challenge Generation', (): void => {
...
+      })
+
+      const challenge = 'Crypto Auth service Yay!'
+      const signer = crypto.createSign('SHA256')
+
+      signer.update(challenge)
+
+      const sign = signer.sign(fakeKeyPair.privateKey, 'base64')
+
+      expect(await verifySign(challenge, sign, realKeyPair.publicKey)).toEqual(false)
```
lewisdaly, line 92: Same as above with `async` and assertions

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,130 @@ describe('Challenge Generation', (): void => {
...
+          type: 'spki', // Key infrasructure
+          format: 'pem' // Encoding format
+        },
+        privateKeyEncoding: {
+          type: 'pkcs8',
+          format: 'pem'
+        }
+      })
+
+      const challenge = 'Crypto Auth service Yay!'
```
akapur99, line 47: not sure but doesn't a valid base 64 string challenge have to end with '='?

akapur99: maybe just also try testing by using `generate` to gen challenges of different sizes (32, 64, etc.) and test verifySign with those?

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,130 @@ describe('Challenge Generation', (): void => {
...
+      expect(await verifySign(challenge, sign, realKeyPair.publicKey)).toEqual(false)
+    })
+  }
+
+  it('verifies correct signature - RSA Key', async (): Promise<void> => {
+    const realKeyPair = crypto.generateKeyPairSync('rsa', {
+      modulusLength: 2048 // Key length in bits
+    })
+
+    const challenge = 'Crypto Auth service Yay!'
```
akapur99, line 101: but also if you're using the same challenge + calling the same functions everytime - consider using beforeAll() beforeEach() in jest to improve code readability

test/unit/lib/challenge.test.ts
akapur99: @ramanmangla does it need to be async at point of call? to just run in the background while other stuff is happening so as not to block

test/unit/lib/challenge.test.ts
akapur99: beforeEach() might also be useful ?

src/lib/challenge.ts
ramanmangla: @lewisdaly that's something I wanted to confirm with you and @eoln . The `crypto` function for` generateChallenge` is promisified to make the overall function async. Should I also make this function aync?

test/unit/lib/challenge.test.ts
ramanmangla: Will implement the change

src/lib/challenge.ts
ramanmangla: Yes, I could do that too.

src/lib/challenge.ts
eoln, line 62:
```suggestion
export function verifySign (
```

src/lib/challenge.ts
```diff
@@ -48,3 +52,21 @@ export async function generate (size: number = 32): Promise<string> {
...
+/*
+ * Helper function to validate signatures using public key
+ * @param challenge Base64 challenge string
+ * @param signature Base64 sign string
+ * @param publicKey Base64 RSA/EC:sec256k1 key string or KeyObject for verification
+ */
+export async function verifySign (
+  challenge: string,
+  signature: string,
+  publicKey: string | KeyObject): Promise<boolean> {
```
eoln, line 65:
```suggestion
  publicKey: string | KeyObject): boolean {
```

test/unit/lib/challenge.test.ts
```diff
@@ -1,4 +1,5 @@
-import { generate } from '../../../src/lib/challenge'
+import { generate, verifySign } from '../../../src/lib/challenge'
+import crypto from 'crypto'
 
 describe('Challenge Generation', (): void => {
   it('Should return a 32 byte string by default', async (): Promise<void> => {
```
eoln, line 5:
```suggestion
  it('Should return a 32 byte string by default', (): void => {
```

test/unit/lib/challenge.test.ts
eoln: if you want to generate `tests` or `it` in a loop, you should also generate proper description, why using magical `4` instead of `7` ?

ramanmangla:
> lgtm - only comment is to add a test which uses generate() to procure a challenge just to see if the modules gel together - cause that's the challenge that this will actually be dealing with

src/lib/challenge.ts
```diff
@@ -19,13 +19,17 @@
...
 
  --------------
  ******/
 
 /* istanbul ignore file */
 
-// Flag to igore BDD testing, which will be dealt with in a later ticket
+/*
+ * Flag to ignore BDD testing, which will be dealt
+ * with in a later ticket
```
lewisdaly, line 31: Please write the ticket number so we can know when to remove this comment

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,173 @@ describe('Challenge Generation', (): void => {
...
+      },
+      privateKeyEncoding: {
+        type: 'pkcs8',
+        format: 'pem'
+      }
+    })
+
+    const sign = signer.sign(keyPair.privateKey, 'base64')
+
+    expect(verifySign(challenge, sign, keyPair.publicKey)).toEqual(true)
```
lewisdaly, line 58: Same comment as earlier. Please separate out the assertions from the test action

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,173 @@ describe('Challenge Generation', (): void => {
     expect(Buffer.byteLength(challenge, 'base64')).toBe(8)
   })
 })
+
+// Each test generates a random key pair
+describe('Signature Verification', (): void => {
```
lewisdaly, line 32: Can you please add a test case for the error that is being thrown 'Unable to verify signature'?

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,173 @@ describe('Challenge Generation', (): void => {
...
+      }
+    })
+
+    const sign = signer.sign(keyPair.privateKey, 'base64')
+
+    expect(verifySign(challenge, sign, keyPair.publicKey)).toEqual(true)
+  })
+
+  // Using another challenge message
+  it('verifies correct signature - EC Key', (): void => {
```
eoln, line 62: there is no need to repeat the test with another challenge, drop it

src/lib/challenge.ts
ramanmangla: @eoln mentioned to me a couple of weeks ago that we only need to reference the ticket number in the PR.

test/unit/lib/challenge.test.ts
ramanmangla: Could you suggest something here @lewisdaly ? I was thinking about this but I didn't know how to get the function to throw an error here.

I just followed the pattern in the existing function from @akapur99 . There is no error test case for that.

src/lib/challenge.ts
```diff
@@ -48,3 +52,26 @@ export async function generate (size: number = 32): Promise<string> {
...
+ * @param signature Base64 sign string
+ * @param publicKey Base64 RSA/EC:secp256k1 key string or KeyObject for verification
+ */
+export function verifySign (
+  challenge: string,
+  signature: string,
+  publicKey: string | crypto.KeyObject): boolean {
+  try {
+    // Hash using SHA256
+    const verifier: crypto.Verify = crypto.createVerify('SHA256')
```
eoln, line 68: there is no need to create verifier for every call of `verifySign` it will slow down verification, make `verifier` module's constant instead.

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,173 @@ describe('Challenge Generation', (): void => {
...
+      privateKeyEncoding: {
+        type: 'pkcs8',
+        format: 'pem'
+      }
+    })
+
+    const sign = signer.sign(keyPair.privateKey, 'base64')
+
+    expect(verifySign(challenge, sign, keyPair.publicKey)).toEqual(true)
+  })
```
eoln, line 59: here you should check that verifier's `update` and `verify` was used properly,

src/lib/challenge.ts
```diff
@@ -48,3 +52,26 @@ export async function generate (size: number = 32): Promise<string> {
     throw error
   }
 }
+
```
eoln, line 55: there is a need to put into system configuration what type of publicKey is used (RSA or EC with additional parameters), also which hash algorithm is used to sign and verify SHA256 - and what else is allowed by fido spec? 
Let don't hardcode these parameters in the code

src/lib/challenge.ts
```diff
@@ -48,3 +52,26 @@ export async function generate (size: number = 32): Promise<string> {
...
+  publicKey: string | crypto.KeyObject): boolean {
+  try {
+    // Hash using SHA256
+    const verifier: crypto.Verify = crypto.createVerify('SHA256')
+
+    verifier.update(challenge)
+
+    return verifier.verify(publicKey, signature, 'base64')
+  } catch (error) {
+    Logger.push(error).error('Unable to verify signature')
```
eoln, line 74: where this line of code is tested?

src/lib/challenge.ts
ramanmangla: @eoln you need a different verifier for each signature. Signing many challenges with the same one doesn't work.

src/lib/challenge.ts
ramanmangla: @eoln The verifier works with many different types of keys without an explicit need to specify which keys it should work with. FIDO allows RSA and EC for some curves. I have tested the function for RSA and the EC I believe we are using in the PISP app.

src/lib/challenge.ts
ramanmangla: Could you suggest something here @eoln ? I was thinking of adding a test for this but I didn't know how to get the function to throw an error here.

I just followed the pattern in the existing function from @akapur99 . There is no error test case for that.

test/unit/lib/challenge.test.ts
jgeewax: Not sure I agree. I think we want to test the behavioral end result not the implementation details of calling specific methods. Let's leave this to simply verify that the challenge checks out as we know it should.

test/unit/lib/challenge.test.ts
jgeewax: I would, however, consider having some hard coded base-64 characters here that represent a public key, challenge, and signature that you know to be right. Then pump those in and verify that all checks out. The goal here is to avoid a scenario where there is a bug in the challenge generation AND the verification.

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +27,173 @@ describe('Challenge Generation', (): void => {
...
+        format: 'pem'
+      }
+    })
+
+    const sign = signer.sign(fakeKeyPair.privateKey, 'base64')
+
+    expect(verifySign(challenge, sign, realKeyPair.publicKey)).toEqual(false)
+  })
+
+  it('returns false on incorrect signature - wrong challenge', (): void => {
```
jgeewax, line 117: Maybe another one with hard coded base-64 bytes also (as noted above)

src/lib/challenge.ts
eoln: @jgeewax there is a need for a designer decision: what type of hash/verify algorithm should we implement?

src/lib/challenge.ts
eoln: bad @akapur99 :-P 
@ramanmangla you should use jest.mock on `crypto` module with mocked function `createVerify` which throws an error so your catch is activated

test/unit/lib/challenge.test.ts
eoln: `verifySign` is returning `true` or `false` for a given set of parameters. We are not implementing crypto algorithm to verify the sign - we are delegating this difficult task to the `crypto` module, so we should test how we are using `crypto` interface.

src/lib/challenge.ts
ramanmangla: @eoln we are accepting many types of keys and using SHA256 for hashing. I have confirmed with with @jgeewax . I could add more documentation on which keys are accepted and which have been tested properly.

eoln: That's great we will support many types of keys and hashing algorithms! 
@jgeewax How the parties will get an agreement which type of keys & algorithms will be used to sing and verify for the given challenge? Maybe the key/hash type should be the part of requests so mobile app will know how to sign, and dfsp/auth-service will know how to verify signature. Then it means that key and hash algorithm types shouldn't be hardcoded in the code but passed via parameters

jgeewax: Just putting in writing what we discussed in stand-up before: the different types of keys should be specified in the key content itself. The choice of what type of key to use (RSA, EC, etc) is up to the device. For now I say we just say we'll use RSA for everything and make sure that works.

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +29,226 @@ describe('Challenge Generation', (): void => {
...
+
+    anotherSigner.update(anotherChallenge)
+
+    const sign = signer.sign(keyPair.private, 'base64')
+    const verified = verifySign(message, sign, keyPair.public)
+
+    expect(verified).toEqual(false)
+  })
+
+  it('properly uses crypto functions and handles exceptions', (): void => {
```
eoln, line 219:
```suggestion
  it('properly uses crypto.createVerify function and handles exceptions', (): void => {
```

test/unit/lib/challenge.test.ts
```diff
@@ -26,3 +29,226 @@ describe('Challenge Generation', (): void => {
...
+    const { message, keyPair } = Credential.RSA
+
+    // Setting up mocks
+    const createVerifySpy = jest.spyOn(crypto, 'createVerify').mockImplementationOnce((): crypto.Verify => {
+      throw new Error('Unable to create Verify in mock')
+    })
+
+    const loggerPushSpy = jest.spyOn(Logger, 'push')
+    const loggerErrorSpy = jest.spyOn(Logger, 'error').mockImplementation((): void => {})
```
eoln, line 228:
```suggestion
    // we mock `Logger.error` because it indirectly calls Logger.push and we want to only test usage of Logger's interface
    const loggerErrorSpy = jest.spyOn(Logger, 'error').mockImplementation((): void => {})
```

src/lib/challenge.ts
```diff
@@ -48,3 +53,30 @@ export async function generate (size: number = 32): Promise<string> {
...
+/**
+ * Helper function to validate signatures using public key
+ * @param challenge UTF-8 challenge string
+ * @param signature Base64 sign string
+ * @param publicKey PEM Base64 Public key string or KeyObject for verification
+ *
+ * Currently, the implementation focuses on RSA 2048 and ECDSA:secp256k1 keys.
+ * Support for additional keys can be extended further.
+ */
+export function verifySign (
```
jgeewax, line 66:
```suggestion
export function verifySignature (
```

PR #19, opened 2020-07-21
Third party request

akapur99:
* Creates `requests` module which imports and instantiates an instance of `ThirdPartyRequests` from `sdk-standard-components`.
* Sets up config files to enable the above.
* Testing will be handled in #11


src/lib/requests.ts
```diff
@@ -0,0 +1,58 @@
...
+ - Raman Mangla <ramanmangla@google.com>
+ - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+
+ --------------
+ ******/
+/* istanbul ignore file */
+// Testing will be covered in PR #11
+
+import { config } from './config'
+// @ts-ignore
```
ramanmangla, line 36: Is this required now?

src/lib/requests.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Logger, ThirdpartyRequests, BaseRequestConfigType } from '@mojaloop/sdk-standard-components'
+
+// Config file to instantiate ThirdPartyRequest object
+const configRequest: BaseRequestConfigType = {
+  dfspId: config.get('PARTICIPANT_ID') as string,
+  logger: new Logger(),
+  // TODO: Decide on below later
```
ramanmangla, line 43: Could you add comments here about whatever the TLS issue was? Maybe open a ticket and reference that here?

ramanmangla: Will we also have to update helm charts and other deployment config?

src/lib/requests.ts
akapur99: what is? the ts ignore? yup

src/lib/requests.ts
akapur99: good catch! will do the former
I'm not sure about the latter - @lewisdaly thoughts?

akapur99, line 43:
```suggestion
  // TODO: Decide on below later - Handled in future ticket #361
```

PR #20, opened 2020-07-22
[WIP] Integration tests Setup

spikerheado1234:
DO NOT BOTHER READING, WANTED TO CHECK IF CI BUILD WAS PASSING!

Sets up integration testing framework, to address ticket #368.


package.json
```diff
@@ -26,14 +26,15 @@
     "test:bdd": "jest --config './jest.bdd.config.js' --testMatch '**/test/step-definitions/**/*.step.ts'",
     "test:coverage": "jest --coverage --coverageThreshold='{}'",
     "test:coverage-check": "jest --coverage",
-    "test:integration": "echo \"test:integration - TODO...\"",
+    "test:integration": "jest --runInBand --testMatch '**/test/integration/**/*.(test|spec).ts'",
```
eoln, line 29:
```suggestion
    "test:integration": "jest --testMatch '**/test/integration/**/*.(test|spec).ts'",
```
why we want to run integration tests in sequence?

src/lib/db.ts
```diff
@@ -24,16 +27,29 @@
...
 
 import Knex from 'knex'
 import Config from '../../config/knexfile'
 import ConsentDB from '../model/consent'
 
-const Db: Knex = Knex(Config.test)
-const consentDB: ConsentDB = new ConsentDB(Db)
+function getKnexInstance (): Knex {
+  let Db: Knex
+  if (process.env.NODE_ENV === 'test') {
```
eoln, line 40: let use `switch` instead of if/else chain

src/cli.ts
```diff
@@ -1,5 +1,8 @@
 #!./node_modules/.bin/ts-node
 
+/* To ignore from integration and BDD testing, addressed in #354 and #368 */
```
eoln, line 3: can somebody explain to me what is the problem with `istanbul` that we need to add `ignore file`

test/integration/index.test.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ * Lewis Daly <lewis@vesselstech.com>
+ * Paweł Marzec <pawel.marzec@modusbox.com>
+ --------------
+ ******/
+
+
+describe('index', (): void => {
```
eoln, line 27: rally? it was the only temp. example, please add some real unit test for index

src/cli.ts
lewisdaly: I think it's because the BDD enforces code coverage. But the issue with all of these ignores, we are ignoring coverage for BDD as well as the unit tests. So perhaps this isn't the best approach.

src/lib/db.ts
```diff
@@ -32,9 +36,34 @@ import Config from '../../config/knexfile'
 import ConsentDB from '../model/consent'
 import ScopeDB from '../model/scope'
 
-const Db: Knex = Knex(Config.test)
-const consentDB: ConsentDB = new ConsentDB(Db)
-const scopeDB: ScopeDB = new ScopeDB(Db)
+function getKnexInstance (): Knex {
+  let Db: Knex
+  switch (process.env.NODE_ENV) {
+    case 'test': {
```
ramanmangla, line 42: What if we directly return from each case instead of breaking and returning in the end?

PR #21, opened 2020-07-22
Update/341: Modify Consent update model to validate against changing existing data

ramanmangla:
In reference to ticket number #341 .

This modifies `Consent` update model with a transaction for the following:
 - Adds validation against updating an existing field for a record
 - Allows `NON-NULL` `credentialStatus` to be updated if it is not `ACTIVE`
 - Adds unit tests to test the same

Main Work Files:
 - `src/model/scope/consent.ts`
 - `test/unit/model/consent.test.ts`

### DIFF
--- a/src/model/consent/consent.ts
+++ b/src/model/consent/consent.ts
@@ -83,15 +83,13 @@ export class ConsentDB {
   }
 
   // Update Consent
-  // No validation against Null or illegal updates in models
+  // Only validating against updating existing fields
   public async update (consent: Consent): Promise<number> {
     // Returns number of updated rows
-    // const updateCount: number = await this
-    //   .Db<Consent>('Consent')
-    //   .where({ id: consent.id })
-    //   .update(consent)
-
-    const updateCount = await this.Db.transaction(async (trx: Knex.Transaction): Promise<number> => {
+    // Transaction to make the update atomic
+    return this.Db.transaction(async (trx): Promise<number> => {
+      // Transaction is rolled back automatically if there is
+      // an error and the returned promise is rejected
       const consents: Consent[] = await trx<Consent>('Consent')
         .select('*')
         .where({ id: consent.id })
@@ -104,12 +102,17 @@ export class ConsentDB {
       const existingConsent = consents[0]
       const updatedConsent: Consent = { ...consent }
 
+      // Cannot update fields once they have been added properly to the DB
       for (const key in consent) {
-        if (existingConsent[key as keyof Consent] !== null) {
+        // Only include the fields which are NULL in the DB
+        // Include credentialStatus even if NON-NULL but ensure that it is not 'ACTIVE'
+        if ((existingConsent[key as keyof Consent] !== null && key !== 'credentialStatus') ||
+            (key === 'credentialStatus' && existingConsent[key as keyof Consent] === 'ACTIVE')) {
           delete updatedConsent[key as keyof Consent]
         }
       }
 
+      // If there are no fields that can be updated
       if (Object.keys(updatedConsent).length === 0) {
         return 0
       }
@@ -118,13 +121,6 @@ export class ConsentDB {
         .where({ id: consent.id })
         .update(updatedConsent)
     })
-
-    // // Ensure that the caller knows that the resource does not exist
-    // if (updateCount === 0) {
-    //   throw new NotFoundError('Consent', consent.id)
-    // }
-
-    return updateCount
   }
 
   // Retrieve Consent by ID (unique)
--- a/src/model/consent/consent.ts
+++ b/src/model/consent/consent.ts
@@ -86,15 +86,43 @@ export class ConsentDB {
   // No validation against Null or illegal updates in models
   public async update (consent: Consent): Promise<number> {
     // Returns number of updated rows
-    const updateCount: number = await this
-      .Db<Consent>('Consent')
-      .where({ id: consent.id })
-      .update(consent)
+    // const updateCount: number = await this
+    //   .Db<Consent>('Consent')
+    //   .where({ id: consent.id })
+    //   .update(consent)
 
-    // Ensure that the caller knows that the resource does not exist
-    if (updateCount === 0) {
-      throw new NotFoundError('Consent', consent.id)
-    }
+    const updateCount = await this.Db.transaction(async (trx: Knex.Transaction): Promise<number> => {
+      const consents: Consent[] = await trx<Consent>('Consent')
+        .select('*')
+        .where({ id: consent.id })
+        .limit(1)
+
+      if (consents.length === 0) {
+        throw new NotFoundError('Consent', consent.id)
+      }
+
+      const existingConsent = consents[0]
+      const updatedConsent: Consent = { ...consent }
+
+      for (const key in consent) {
+        if (existingConsent[key as keyof Consent] !== null) {
+          delete updatedConsent[key as keyof Consent]
+        }
+      }
+
+      if (Object.keys(updatedConsent).length === 0) {
+        return 0
+      }
+
+      return trx<Consent>('Consent')
+        .where({ id: consent.id })
+        .update(updatedConsent)
+    })
+
+    // // Ensure that the caller knows that the resource does not exist
+    // if (updateCount === 0) {
+    //   throw new NotFoundError('Consent', consent.id)
+    // }
 
     return updateCount
   }
--- a/src/model/consent/consent.ts
+++ b/src/model/consent/consent.ts
@@ -83,7 +83,6 @@ export class ConsentDB {
   }
 
   // Update Consent
-  // Only validating against updating existing fields
   public async update (consent: Consent): Promise<number> {
     // Returns number of updated rows
     // Transaction to make the update atomic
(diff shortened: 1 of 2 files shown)


src/model/consent/consent.ts
```diff
@@ -83,20 +83,44 @@ export class ConsentDB {
...
-
-    return updateCount
+    // Transaction to make the update atomic
+    return this.Db.transaction(async (trx): Promise<number> => {
+      // Transaction is rolled back automatically if there is
+      // an error and the returned promise is rejected
+      const consents: Consent[] = await trx<Consent>('Consent')
+        .select('*')
+        .where({ id: consent.id })
+        .limit(1)
```
akapur99, line 96: why is this limit necessary? there should only be one consent for a given ID right?

src/model/consent/consent.ts
```diff
@@ -83,20 +83,44 @@ export class ConsentDB {
...
+
+      if (consents.length === 0) {
+        throw new NotFoundError('Consent', consent.id)
+      }
+
+      const existingConsent = consents[0]
+      const updatedConsent: Consent = { ...consent }
+
+      // Cannot update fields once they have been added properly to the DB
```
akapur99, line 105: what do you mean added properly?

src/model/consent/consent.ts
ramanmangla: Yes, it is just good practice to enforce a limit constraint.

src/model/consent/consent.ts
ramanmangla: When fields are at their final or "proper" values. This mainly concerns `credentialStatus` which has a slightly different constraint.

src/model/consent/consent.ts
```diff
@@ -83,20 +83,44 @@ export class ConsentDB {
...
+      for (const key in consent) {
+        // Only include the fields which are NULL in the DB
```
eoln, line 107:
```suggestion
        // Only prepare for update the fields which are NULL in the DB
```

src/model/consent/consent.ts
```diff
@@ -83,20 +83,44 @@ export class ConsentDB {
...
+        // Include credentialStatus even if NON-NULL but ensure that it is not 'ACTIVE'
+        if ((existingConsent[key as keyof Consent] !== null && key !== 'credentialStatus') ||
+            (key === 'credentialStatus' && existingConsent[key as keyof Consent] === 'ACTIVE')) {
```
eoln, line 110:
```suggestion
            (key === 'credentialStatus' && existingConsent.credentialStatus === 'ACTIVE')) {
```

src/model/consent/consent.ts
```diff
@@ -83,20 +83,44 @@ export class ConsentDB {
   }
 
   // Update Consent
-  // No validation against Null or illegal updates in models
+  // Only validating against updating existing fields
   public async update (consent: Consent): Promise<number> {
```
eoln, line 87: maybe we should consider returning `boolean` value instead of `number` and be consistent with the way we do for `insert` method?

ramanmangla: @eoln we return the `updateCount` here is a useful check of how many rows have been changed. The reason `insert` returns a `boolean` is because MySQL and SQLite don't return anything useful on `insert`. So we just had to go with a `boolean`.

eoln: hmm, but we are using limit in transaction context, so after an update, the update count should be always 1 - am I right?

ramanmangla: Yes, it should be 1 always because `Consent` is unique. But the `update` on MySQL and SQLite returns number of rows affected. This also ensures coherence with `Scope` where `update` can return any number of affected rows.

eoln: don't get me wrong @ramanmangla but your code is inside transaction -> it means all atomic SQL commands are totally isolated. So if you do a select in L93 with and there is a uniqueness constraint on relation Consent-Scope you will always get one item, also you will always update one item in L120 ->the transaction guarantee that, so your code will always return 1, so maybe it is good to consider using boolean instead which will be the simpler concept.

eoln: but maybe it is better to keep interface consistency with Scopes.update, so leave it as it is ;-)

ramanmangla: I agree with you. Yes, we can leave it for now for consistency purposes.

src/model/consent/consent.ts
```diff
@@ -83,20 +83,44 @@ export class ConsentDB {
...
+
+      // Cannot update fields once they have been added properly to the DB
+      for (const key in consent) {
+        // Only include the fields which are NULL in the DB
+        // Include credentialStatus even if NON-NULL but ensure that it is not 'ACTIVE'
+        if ((existingConsent[key as keyof Consent] !== null && key !== 'credentialStatus') ||
+            (key === 'credentialStatus' && existingConsent[key as keyof Consent] === 'ACTIVE')) {
+          delete updatedConsent[key as keyof Consent]
+        }
+      }
```
lewisdaly, line 113: There's a lot of hard to read logic in this for statement contained in the `if statement`.

My preference would be to use a JS forEach with a function to build up the `updatedFields` object, for example:

```ts
const updatedFields: { [index: string]: string } = {};
Object.keys(consent).forEach(key => {
  const value = existingConsent[key as keyof Consent]

  // Don't allow overwriting Credential status
  if (key === 'credentialStatus' && value === 'ACTIVE') {
    return;
  }

  // Don't allow overwriting non-null fields
  if (value !== null) {
    return;
  }

  // Ensure type safety - the only fields we wish to update should be strings
  if(typeof value !== 'string') {
    return;
  }

  updatedFields[key] = value;
})
```

> (I don't know if I got the logic right, but hopefully you get the idea)

test/unit/model/consent.test.ts
```diff
@@ -172,7 +193,71 @@ describe('src/model/consent', (): void => {
...
+      const consents: Consent[] = await Db<Consent>('Consent')
+        .select('*')
+        .where({
+          id: completeConsent.id
+        })
+
+      expect(consents[0].createdAt).toEqual(expect.any(String))
+
+      // Conflicting fields (initiatorId and participantId) are still the same
+      expect(consents[0]).toEqual(expect.objectContaining(partialConsent))
```
lewisdaly, line 219: Take a look at `https://jestjs.io/docs/en/expect#expectobjectcontainingobject` to give you more flexibility in checking your test assertions.

src/model/consent/consent.ts
ramanmangla: @lewisdaly Isn't `forEach` slower than a `for loop` and the older way to do this? Also, `updatedFields` would be of type `Object` instead of `Consent` and we would have to typecast it to `Consent`.

src/model/consent/consent.ts
ramanmangla: Done

src/model/consent/consent.ts
ramanmangla: Also, the type validation is provided by TS and will be a responsibility of the handler.

eoln: @ramanmangla forEach is slower, but the real difference will be visible for more than hundreds of items on the list. @lewisdaly 's code is much cleaner! I would suggest to use it. Type validation is only done at compile-time, at execution -time all evils are waiting for your soul.

lewisdaly: Indeed I agree with @eoln's agreeing to me. By passing in a function that gets evaluated over every key of the object, we get nice and easy ways to `return` early in case we don't want do anything, which is much easier to read than `deleting` in my opinion.

eoln: is it called mutual agreement @lewisdaly

eoln: but for loop, you can use the `continue` statement in the same you are using `return` in callback

ramanmangla: I definitely agree that the code would be much cleaner. I will make the changes.

src/model/consent/consent.ts
```diff
@@ -83,20 +83,51 @@ export class ConsentDB {
...
+        // Cannot overwrite an `ACTIVE` credentialStatus
+        if (key === 'credentialStatus' && value === 'ACTIVE') {
+          return
+        }
+
+        // Cannot overwrite non-null fields
+        if (value !== null && key !== 'credentialStatus') {
+          return
+        }
+
```
eoln, line 118: @ramanmangla It is okay as it is now. But, if you would like to have some fun at the job, as a programming challenge, I would refactor this logic using `filter` and `reduce` instead `forEach`.

test/unit/model/consent.test.ts
```diff
@@ -172,7 +193,64 @@ describe('src/model/consent', (): void => {
 
       expect(consents[0].id).toEqual(partialConsent.id)
       expect(consents[0].createdAt).toEqual(expect.any(String))
-      expect(consents[0]).toEqual(expect.objectContaining(completeConsent))
+      expect(consents[0]).toEqual(expect.objectContaining(consentWithOnlyUpdateFields))
+    })
+
+    it('updates existing consent with only non-conflicting fields from a consent', async (): Promise<void> => {
```
eoln, line 199: you should explain here, maybe in comments, or better in code, the meaning of `conflicting fields`

test/unit/model/consent.test.ts
```diff
@@ -60,6 +60,27 @@ const completeConsent: Consent = {
...
+  credentialType: 'FIDO',
+  credentialStatus: 'PENDING',
+  credentialChallenge: 'xyhdushsoa82w92mzs',
+  credentialPayload: 'dwuduwd&e2idjoj0w'
+}
+
+const conflictingConsent: Consent = {
+  id: '1234',
+  // Conflicting initiatorId and participantId
+  initiatorId: 'pisp-0000-1133',
```
eoln, line 75: does it mean that `conflictingConset` has the same id as `completeConsent` but different `initiatorId` and `participantId` ? If yes please explain it in comments and code

test/unit/model/consent.test.ts
```diff
@@ -60,6 +60,27 @@ const completeConsent: Consent = {
...
+  credentialId: '123',
+  credentialType: 'FIDO',
+  credentialStatus: 'PENDING',
+  credentialChallenge: 'xyhdushsoa82w92mzs',
+  credentialPayload: 'dwuduwd&e2idjoj0w'
+}
+
+const conflictingConsent: Consent = {
+  id: '1234',
+  // Conflicting initiatorId and participantId
```
eoln, line 74: does it mean that `conflictingConset` has the same `id` as `completeConsent` but different `initiatorId` and `participantId` ? Please explain how to understand the `conflict`.

ramanmangla: Yes.

src/model/consent/consent.ts
```diff
@@ -83,20 +83,51 @@ export class ConsentDB {
...
-      .update(consent)
-
-    // Ensure that the caller knows that the resource does not exist
-    if (updateCount === 0) {
-      throw new NotFoundError('Consent', consent.id)
-    }
-
-    return updateCount
+    // Transaction to make the update atomic
+    return this.Db.transaction(async (trx): Promise<number> => {
```
jgeewax, line 89: Generally, I think shorthand for this is `txn`.

src/model/consent/consent.ts
```diff
@@ -83,20 +83,51 @@ export class ConsentDB {
...
+      // Transaction is rolled back automatically if there is
+      // an error and the returned promise is rejected
+      const consents: Consent[] = await trx<Consent>('Consent')
+        .select('*')
+        .where({ id: consent.id })
+        .limit(1)
```
jgeewax, line 95: This isn't necessary, right?

src/model/consent/consent.ts
ramanmangla: I was following these links which use `trx`:
 - http://knexjs.org/#Transactions
 - https://vincit.github.io/objection.js/guide/transactions.html

src/model/consent/consent.ts
ramanmangla: You suggested having a limit for `select` with `Consent` which is a unique resource. That is why we also do the same in `retrieve`.

PR #26, opened 2020-07-30
384: Unlinking 

akapur99:
Ticket 384 on Zenhub
https://app.zenhub.com/workspaces/pisp-5e8457b05580fb04a7fd4878/issues/mojaloop/mojaloop/384

- [x] Add `status, revokedAt` property to Consent model, migrations, seeds
- [x] Enforce *no-modification* rule for revoked consent
- [x] Refactor/Fix all previously implemented handlers/code to be compliant with the new Consent Schema
- [x] Request handler for `POST /consents/{id}/revoke`
- [x] Validation of request
- [x] Return 202 Accepted in response
- [x] Revoking consent (DO NOT DELETE)
- [x] New outgoing call `PATCH /consents/123 { status: REVOKED, RevokedAt: timestamp }` to the PISP - sending back the 2 updated consent properties
- [x] Unit Tests for above
- [x] Update definitions.yaml, swagger.json and other documentation if needed


src/server/handlers/consents/{ID}/revoke.ts
```diff
@@ -0,0 +1,99 @@
...
+      Logger.push(error)
+      Logger.error('Error in retrieving consent')
+
+      // If consent cannot be retrieved using given ID, send PUT ...error back
+      // TODO: Error Handling dealt with in future ticket #355
+      throw (new Error('NotImplementedYetError'))
+    }
+
+    // If request is not intiated by valid source, send PUT ...error back
+    if (!isConsentRequestInitiatedByValidSource(consent, request)) {
```
ramanmangla, line 67: I think this is handled by the OpenAPI-Hapi plugin.

src/server/handlers/consents/{ID}/revoke.ts
```diff
@@ -0,0 +1,99 @@
...
+      // TODO: Error Handling dealt with in future ticket #355
+      throw (new Error('NotImplementedYetError'))
+    }
+
+    // If Consent is ACTIVE, revoke it and update database
+    if (consent.status === 'ACTIVE') {
```
ramanmangla, line 73: Maybe we could throw an error if the `status` is not active. Maybe `BAD REQUEST`?

src/model/consent/consent.ts
```diff
@@ -52,12 +52,14 @@ export interface Consent {
...
   initiatorId?: string;
   participantId?: string;
+  status: string;
   credentialId?: string;
   credentialType?: string;
   credentialStatus?: string;
   credentialPayload?: string;
   credentialChallenge?: string;
   createdAt?: Date;
+  revokedAt?: string;
```
ramanmangla, line 62: Should be a Date.

src/interface/definitions.yaml
```diff
@@ -50,13 +50,26 @@ Consent:
...
+ConsentStatusEnum:
+  title: ConsentStatus
+  type: string
+  enum:
+    - ACTIVE
+    - REVOKED
+  description: |
+    The status of the Credential's creation
+    - "ACTIVE" - The Consent is valid, and ready to be used by the PISP
+    - "REVOKED" - The Consent is no longer valid and has been requested to be deleted
```
ramanmangla, line 71: Maybe `The Consent is no longer valid and has been revoked` is better?

src/model/consent/consent.ts
akapur99: yeah lets chat about this on monday

src/server/handlers/consents/{ID}/revoke.ts
akapur99: nope

src/server/handlers/consents/{ID}/revoke.ts
akapur99: do you mean response? cant do that because we sent back the ACCEPTED code already

src/interface/definitions.yaml
akapur99: makes sense

src/domain/consents/revoke.ts
```diff
@@ -42,13 +42,21 @@ import { Consent } from '../../model/consent'
...
 
+/**
+ * Validates if the request is initiated by a valid source
+ * Compares Consent object's initiator ID with source
+ */
 export function isConsentRequestInitiatedByValidSource (
   consent: Consent,
   request: Request): boolean {
   const fspiopSource = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
   return (consent && consent.initiatorId === fspiopSource)
```
eoln, line 52:
```suggestion
  return (consent?.initiatorId === fspiopSource)
```

src/domain/consents/revoke.ts
```diff
@@ -42,13 +42,21 @@ import { Consent } from '../../model/consent'
...
 }
 
+/**
+ * Revoke status of consent object, update in the database
+ * and return consent
+ */
 export async function revokeConsentStatus (
```
eoln, line 59: there is no need for async here

src/domain/consents/revoke.ts
```diff
@@ -57,6 +65,9 @@ export async function revokeConsentStatus (
   return consent
```
eoln, line 64: you shouldn't update Consent if it is already REVOKED and where is an update in DB invocation? you should add guard if(consent.status !== 'ACTIVE') { throw ... }

src/server/handlers/consents/{ID}/revoke.ts
```diff
@@ -45,6 +45,10 @@ import { Enum } from '@mojaloop/central-services-shared'
 import { consentDB } from '../../../../lib/db'
 import { Consent } from '../../../../model/consent'
 
+/**
+ * Asynchronously deals with validating request, revoking consent object
+ * and send outgoing PATCH consent/{id}/revoke request to switch
+ */
 export async function validateRequestAndRevokeConsent (
   request: Request): Promise<void> {
   const consentId = request.params.id
```
eoln, line 55: in L77 remove `if` guard which should be implemented in `revokeConsentStatus` and properly catch thrown exception using PUT ../error

src/domain/consents/revoke.ts
akapur99: because we're `await`ing the async update

src/domain/consents/revoke.ts
akapur99: so at the  moment we're not updating consent if it's already revoked. we just skip the updating/revoking and send the PATCH request directly - since there's a `revokedAt` field the assumption is the PISP/DFSP can see that
Lmk if that seems right

src/domain/consents/revoke.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Revoke status of consent object, update in the database
+ * and return consent
+ */
+export async function revokeConsentStatus (
+  consent: Consent): Promise<Consent> {
+  if (consent.status !== 'ACTIVE') {
+    // TODO: Confirm this is how we want to handle this
+    return consent
+    // throw (new Error('Consent already revoked'))
```
eoln, line 64: let `throw new CantRevokeConsentError` returning `consent` here we are silently doing nothing - the client should be informed explicitly

src/domain/consents/revoke.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+  consent.status = 'REVOKED'
+  consent.revokedAt = Date.prototype.toISOString()
+  await consentDB.update(consent)
+  return consent
+}
+
+/**
+ * Send outgoing PATCH consent/{id}/revoke request
```
eoln, line 73: this comment is misinforming about function behaviour

src/interface/swagger.json
```diff
@@ -634,6 +634,12 @@
           ],
           "description": "PISP identifier who uses this Consent"
         },
+        "status": {
```
eoln, line 637: swagger.json should be generated using swagger-cli from api-template

src/server/handlers/onValidateFail.ts
```diff
@@ -25,6 +31,10 @@
 import Boom from '@hapi/boom'
 import { Request, Lifecycle, ResponseToolkit } from '@hapi/hapi'
 
-export default function onValidateFail (_request: Request, _h: ResponseToolkit, err?: Error | undefined): Lifecycle.ReturnValue {
+export default function onValidateFail (
+  _request: Request,
+  _h: ResponseToolkit,
+  err?: Error | undefined
```
eoln, line 37:
```suggestion
  err?: Error
```

test/unit/server/handlers/consents/{ID}/generateChallenge.test.ts
```diff
@@ -225,6 +238,24 @@ describe('server/handlers/consents/{ID}/generateChallenge', (): void => {
         expect(mockUpdateConsentCredential).not.toHaveBeenCalled()
       })
 
+    it('Should throw an error due revoked consent', async (): Promise<void> => {
+      mockConsentDbRetrieve.mockResolvedValueOnce(completeConsentRevoked)
```
eoln, line 242: mixing completeConsentRevoked with partialConsent?

test/unit/server/handlers/consents/{ID}/revoke.test.ts
```diff
@@ -0,0 +1,256 @@
...
+        accountId: '232345',
+        actions: ['acc.accessSaving']
+      }
+    ],
+    credential: null
+  }
+}
+
+// @ts-ignore
+const h: ResponseToolkit = {
```
eoln, line 80: broken mocking -> code function should return ResponseObject not number!

test/unit/server/handlers/consents/{ID}/revoke.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      mockValidateRequestAndRevokeConsent.mockResolvedValue()
+    })
+
+    it('Should return 202 success code',
+      async (): Promise<void> => {
+        const response = await Handler.post(
+          request as Request,
+          h as ResponseToolkit
+        )
+        expect(response).toBe(202)
```
eoln, line 253: this is false positive test -> @akapur99 tell me why?

src/interface/swagger.json
akapur99: not completely sure how to do that

src/domain/consents/revoke.ts
akapur99: good catch sorry!

src/domain/consents/revoke.ts
akapur99: Agreed but i think this is something that requires discussion 
do we want to inform explicitly?  https://github.com/mojaloop/auth-service/pull/26#discussion_r466385862
as mentioned in this discussion^

test/unit/server/handlers/consents/{ID}/generateChallenge.test.ts
akapur99: ah so i've kept all the consent ID's the same - but I see that is not providing for good readability - will modify

test/unit/server/handlers/consents/{ID}/revoke.test.ts
akapur99: wasn't really sure what to test in the main handler - just need to confirm that as long background function is called - handler resolves

eoln: @akapur99 NOPE! it is false positive because the mock in L80 is invalid

akapur99:
@ramanmangla please help figure out what's going on with integration tests and the connection refusal?

src/domain/consents/revoke.ts
```diff
@@ -0,0 +1,84 @@
...
+  return (consent?.initiatorId === fspiopSource)
+}
+
+/**
+ * Revoke status of consent object, update in the database
+ * and return consent
+ */
+export async function revokeConsentStatus (
+  consent: Consent): Promise<Consent> {
+  if (consent.status !== 'ACTIVE') {
```
akapur99, line 61:
```suggestion
  if (consent.status === 'REVOKED') {
```

src/domain/consents/revoke.ts
akapur99: decided to add logger note - for this 
but continue remaining silent since we are returning `revokedAt` timestamp

migrations/20200624121732_consent.ts
```diff
@@ -28,17 +29,21 @@ export async function up (knex: Knex): Promise<void | Knex.SchemaBuilder> {
...
-        })
+        return knex.schema.createTable('Consent',
+          (t: Knex.CreateTableBuilder): void => {
+            // TODO: Confirm string length for status and revoked at
+            t.string('id', 32).primary().notNullable()
+            t.string('status', 32).notNullable()
+            t.string('initiatorId', 32).notNullable()
+            t.string('participantId', 32).notNullable()
+            t.timestamp('createdAt').defaultTo(knex.fn.now())
+            t.string('revokedAt', 256).nullable()
```
ramanmangla, line 40: Add a comment here about the `ISO string  Date` which might be easier to identify future changes.

src/domain/consents/generateChallenge.ts
```diff
@@ -54,7 +54,7 @@ export function isConsentRequestInitiatedByValidSource (
   request: Request,
   consent: Consent): boolean {
   const fspiopSource = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
-  return (consent && consent.initiatorId === fspiopSource)
+  return (consent?.initiatorId === fspiopSource)
```
ramanmangla, line 57: What happens when this is compiled to JS?

test/integration/model/consent.test.ts
```diff
@@ -237,13 +245,14 @@ describe('src/model/consent', (): void => {
...
           // Even other fields are the same
           ...completeConsent,
           createdAt: expect.any(Date),
           // credentialStatus is updated to 'ACTIVE'
-          credentialStatus: 'ACTIVE'
+          credentialStatus: 'ACTIVE',
+          revokedAt: null
         }
 
         expect(consents[0]).toEqual(expectedConsent)
```
ramanmangla, line 258: Add integration tests to check for correct update/insert behavior of `revokedAt`.

test/unit/model/consent.test.ts
```diff
@@ -315,53 +333,55 @@ describe('src/model/consent', (): void => {
...
+          .where({ id: completeConsent.id })
 
-      expect(consents.length).toEqual(0)
-      expect(scopes.length).toEqual(0)
-    })
+        expect(consents.length).toEqual(0)
+        expect(scopes.length).toEqual(0)
+      })
   })
 })
```
ramanmangla, line 387: Add unit tests for correct insert/update behavior of `revokedAt`.

akapur99: Separation of concerns - will make a separate PR for further testing

test/integration/model/consent.test.ts
akapur99: Separation of concerns - will make a separate PR for further testing

migrations/20200624121732_consent.ts
akapur99: makes sense

src/domain/consents/generateChallenge.ts
akapur99: not sure but ask pawel?

migrations/20200624121732_consent.ts
akapur99, line 40:
```suggestion
            // TODO: Do we need to change internal representation to Date object
            t.string('revokedAt', 256).nullable()
```

src/domain/consents/generateChallenge.ts
jgeewax, line 57: If consent is null AND fsoiopSource is null, this should evaluate to `true` which is probably not what you want.... You should probably add a test to verify this before changing it back.

src/domain/consents/generateChallenge.ts
```diff
@@ -54,7 +54,7 @@ export function isConsentRequestInitiatedByValidSource (
   request: Request,
   consent: Consent): boolean {
   const fspiopSource = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
```
jgeewax, line 56:
```suggestion
  const fspiopSource = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
  if (fspiopSource === null) return false;
```

jgeewax: You may also want to check for empty string values here.

src/domain/consents/revoke.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Consent } from '../../model/consent'
+import { Enum } from '@mojaloop/central-services-shared'
+import Logger from '@mojaloop/central-services-logger'
+import SDKStandardComponents from '@mojaloop/sdk-standard-components'
+
+/**
+ * Validates if the request is initiated by a valid source
+ * Compares Consent object's initiator ID with source
+ */
+export function isConsentRequestInitiatedByValidSource (
```
jgeewax, line 49: We should factor this out. Ask Lewis or @eoln  where it should live.

src/domain/consents/revoke.ts
```diff
@@ -0,0 +1,84 @@
...
+  consent.status = 'REVOKED'
+  consent.revokedAt = (new Date()).toISOString()
+  await consentDB.update(consent)
+  return consent
+}
+
+/**
+ * Generate outgoing PATCH consent/{id}/revoke request body
+ */
+export function generatePatchConsentRequest (
```
jgeewax, line 75:
```suggestion
export function generatePatchRevokedConsentRequest (
```

src/domain/consents/revoke.ts
```diff
@@ -0,0 +1,84 @@
...
+  consent: Consent
+): SDKStandardComponents.PatchConsentsRequest {
+  const requestBody: SDKStandardComponents.PatchConsentsRequest = {
+    status: consent.status as string as 'REVOKED',
```
jgeewax, line 79:
```suggestion
    status: 'REVOKED',
```

src/domain/consents/revoke.ts
```diff
@@ -0,0 +1,84 @@
...
+  return consent
+}
+
+/**
+ * Generate outgoing PATCH consent/{id}/revoke request body
+ */
+export function generatePatchConsentRequest (
+  consent: Consent
+): SDKStandardComponents.PatchConsentsRequest {
+  const requestBody: SDKStandardComponents.PatchConsentsRequest = {
```
jgeewax, line 78:
```suggestion
  if (consent.status !== 'REVOKED')
    throw new Error('Attempting to generate request for non-revoked consent!')
  const requestBody: SDKStandardComponents.PatchConsentsRequest = {
```

src/model/consent/consent.ts
```diff
@@ -121,8 +123,13 @@ export class ConsentDB {
           return
         }
 
+        // Cannot overwrite REVOKED status Consent
```
jgeewax, line 126:
```suggestion
        // Once revoked, Consent status should never be updated.
```

src/server/handlers/consents/{ID}/revoke.ts
```diff
@@ -0,0 +1,107 @@
...
+      throw (new Error('NotImplementedYetError'))
+    }
+
+    // If request is not intiated by valid source, send PUT ...error back
+    if (!isConsentRequestInitiatedByValidSource(consent, request)) {
+      // TODO: Error Handling dealt with in future ticket #355
+      throw (new Error('NotImplementedYetError'))
+    }
+
+    // If Consent is ACTIVE, revoke it and update database
```
jgeewax, line 77:
```suggestion
    // If Consent is ACTIVE, revoke it and update database. If already revoked, leave it alone but don't throw an error.
```

ambient.d.ts
```diff
@@ -33,3 +33,5 @@ declare module '@mojaloop/central-services-shared'
 declare module '@hapi/good'
 declare module 'hapi-openapi'
 declare module 'blipp'
+
+declare module 'env-variable'
```
akapur99, line 37:
```suggestion
```

src/domain/consents/revoke.ts
eoln: I think `validators.ts` would be a good place to keep all validation related code and then use `validators.` prefix
```
import validators from './validators.ts'
.....
if( validators.isConsentRequestInitiatedByValidSource(.....) ) { .... }
```
so code will be more readable.... if the names of validators will be reasonably short ;-)

akapur99: makes sense - so `validators.ts` would live in domain folder? 
Was talking to @lewisdaly  about possible having it live in the `shared` or `lib` folder?

eoln: I wouldn't keep them in `shared` which is dedicated to low-level tools. Let `validators.ts` live in `/src/domain/validators.ts`

src/server/handlers/consents/{ID}/generateChallenge.ts
```diff
@@ -79,6 +79,14 @@ export async function generateChallengeAndPutConsent (
       throw (new Error('NotImplementedYetError'))
     }
 
+    // Revoked consent should NOT be touched.
+    if (consent.status === 'REVOKED') {
+      // TODO: Confirm what to do here
+      // Error Handling dealt with in future ticket #355
+      // send PUT ...error back ?
+      throw (new Error('Revoked Consent'))
```
eoln, line 87: yes, send PUT error

src/server/handlers/consents/{ID}/revoke.ts
```diff
@@ -0,0 +1,107 @@
...
+    // If request is not intiated by valid source, send PUT ...error back
+    if (!isConsentRequestInitiatedByValidSource(consent, request)) {
+      // TODO: Error Handling dealt with in future ticket #355
+      throw (new Error('NotImplementedYetError'))
+    }
+
+    // If Consent is ACTIVE, revoke it and update database. If already revoked, leave it alone but don't throw an error.
+    consent = await revokeConsentStatus(consent)
+
+    // Outgoing call to PUT consents/{ID}/revoke
```
eoln, line 80:
```suggestion
    // Outgoing call to PATCH consents/{ID}/revoke
```

test/integration/seeds/consent.test.ts
```diff
@@ -49,31 +49,37 @@ describe('testing Consent table', (): void => {
       id: '123',
       initiatorId: 'PISPA',
       participantId: 'DFSPA',
+      status: 'ACTIVE',
```
eoln, line 52: @akapur99 shouldn't we reseed also some data with REVOKED status to have data ready for tests of `revoke` related logic?

src/model/consent/consent.ts
```diff
@@ -108,6 +110,11 @@ export class ConsentDB {
         throw new NotFoundError('Consent', consent.id)
       }
 
+      // Cannot overwrite REVOKED status Consent
```
eoln, line 113: This is business logic - shouldn't it be in the domain part of code. @jgeewax

test/unit/domain/consents/revoke.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    it('Should throw an error as request headers are missing', (): void => {
+      expect((): void => {
+        isConsentRequestInitiatedByValidSource(
+          partialConsentActive, requestNoHeaders as Request)
+      }).toThrowError()
+    })
+  })
+
+  describe('revokeConsentStatus', (): void => {
+    it('Should return a revoked consent', async (): Promise<void> => {
```
eoln, line 146: if there is a need to have separate test cases for `partialConsent` and `completeConsent` there should be some comment here why. Assertions are the same and difference is only exposed in parameter to `revokeConsentStatus` and describe description isn't helping to understand. Improve readability, please

test/unit/domain/consents/revoke.test.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    it('Should also return a revoked consent', async (): Promise<void> => {
+      const revokedConsent = await revokeConsentStatus(completeConsentActive)
+      expect(revokedConsent.status).toBe('REVOKED')
+      expect(revokedConsent.revokedAt).toBeDefined()
+      expect(mockConsentUpdate).toHaveBeenCalled()
+      expect(mockLoggerPush).not.toHaveBeenCalled()
+    })
+
+    it('Should return the consent without any operations, if already revoked',
```
eoln, line 162: does consent have any operations? The `it` description needs clarity.

test/unit/domain/consents/revoke.test.ts
```diff
@@ -0,0 +1,229 @@
...
+
+        await expect(revokeConsentStatus(partialConsentActive2))
+          .rejects
+          .toThrowError('Test Error')
+
+        expect(mockConsentUpdate).toHaveBeenCalled()
+        expect(mockLoggerPush).not.toHaveBeenCalled()
+      })
+  })
+
```
eoln, line 182: there is a lack of tests for cases when `revokeConsentStatus` is called with  REVOKED Consent arg.

test/unit/domain/consents/revoke.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        .toStrictEqual(requestBody)
+    })
+
+    it('Should throw an error as consent is null value', (): void => {
+      expect((): void => {
+        generatePatchConsentRequest(
+          null as unknown as Consent)
+      }).toThrow()
+    })
+  })
```
eoln, line 209: there is a lack of tests where `generatePatchConsentRequest` is called with ACTIVE Consent arg.

test/unit/server/handlers/consents/{ID}/revoke.test.ts
```diff
@@ -0,0 +1,261 @@
...
+          .rejects.toThrowError('NotImplementedYetError')
+
+        expect(mockConsentRetrieve).toBeCalledWith(consentId)
+        expect(mockIsConsentRequestValid).toBeCalledWith(partialConsentActive, request)
+        expect(mockRevokeConsentStatus).not.toBeCalled()
+        expect(mockGeneratePatchConsentRequest).not.toBeCalled()
+        expect(mockPatchConsents).not.toBeCalled()
+      })
+
+    it('Should throw an error as patchConsents() throws an error',
```
eoln, line 214:
```suggestion
    it('Should propagate errors from patchConsents()',
```
there is a need to improve the readability of `it` descriptions here and around

src/model/consent/consent.ts
jgeewax: I agree - - however this is one of those cases where we previously decided that we don't want the model to be able to do something potentially dangerous (see 120/127 where we ensure we don't overwrite ACTIVE credentialStatus)

eoln (approved):

PR #27, opened 2020-08-03
feat(commits): added linter to enforce conventional commit messages

kleyow:
Opening up the discussion here.

I've included the default commitlint config. Any opinionated changes that anyone suggested be changed?

Only one I have is subject start always be capitalized. (I'm not tied to any one styling as long as it's wide spread adopted)


commitlint.config.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
+  parserPreset: 'conventional-changelog-conventionalcommits',
+  rules: {
+    'body-leading-blank': [1, 'always'],
+    'body-max-line-length': [2, 'always', 100],
+    'footer-leading-blank': [1, 'always'],
+    'footer-max-line-length': [2, 'always', 100],
+    'header-max-length': [2, 'always', 100],
```
akapur99, line 8: Shouldn't max length of header be 50 characters?

kleyow: Typically yes we aim for 50. Depending on who you ask 50 is a guideline. 
It's up to us to decide how we want to move forward

eoln: I would keep 100, using conventional-commits the messages are much longer

commitlint.config.js
```diff
@@ -0,0 +1,37 @@
...
+      ['sentence-case', 'start-case', 'pascal-case', 'upper-case']
+    ],
+    'subject-empty': [2, 'never'],
+    'subject-full-stop': [2, 'never', '.'],
+    'type-case': [2, 'always', 'lower-case'],
+    'type-empty': [2, 'never'],
+    'type-enum': [
+      2,
+      'always',
+      [
```
eoln, line 22: I would import this list from .versionrc file

kleyow: done

commitlint.config.js
kleyow: sounds good

.eslintignore
```diff
@@ -1,3 +1,4 @@
 node_modules/**/*.js
 coverage/*
-.circleci/*
\ No newline at end of file
+.circleci/*
+commitlint.config.js
```
eoln, line 4: I think we need this file linted

commitlint.config.js
```diff
@@ -1,3 +1,9 @@
+const versionrc = require('./.versionrc.js')
+
+const types = versionrc.types.map(function(typeDict) {
```
eoln, line 3:
```suggestion
const types = versionrc.types.map((typeDict) => typeDict.type)
```

kleyow: done

.eslintignore
kleyow: yep. don't know what I was thinking. 
the right thing was to add "typescript/no-var-requires" to eslint file
done.

kleyow:
@eoln ping :D

PR #28, opened 2020-08-04
feat(circleci): add steps to circleci to automate github release

kleyow:
Pretty sure this won't work on the first pass lol...

Just copied the changes over from `rancher-scaler` and I'll increment from there til it's working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ version: 2.1
 ##
 orbs:
   anchore: anchore/anchore-engine@1.6.0
-  deploy-kube: mojaloop/deployment@0.1.6
+  github-release: h-matsuo/github-release@0.1.3
   slack: circleci/slack@3.4.2
 
 ##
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -398,99 +407,151 @@ workflows:
   version: 2
   build_and_test:
     jobs:
+      - build-local:
```
eoln, line 410: why we are building a docker image for every branch? maybe let add some comments with an explanation

lewisdaly: Yeah this was an idea I had so that any tests inside of the docker image can be run along with the unit tests.

.circleci/config.yml
```diff
@@ -398,99 +407,151 @@ workflows:
...
+            - image-scan
+            - test-unit
+            - test-coverage
           filters:
             tags:
-              only: /v[0-9]+(\.[0-9]+)*(\-snapshot)?(\-hotfix(\.[0-9]+))?/
+              only: /.*/
             branches:
               ignore:
                 - /.*/
```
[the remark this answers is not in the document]
lewisdaly: Yes, but circleci doesn't see `tags` as having a branch.

kleyow:
Closed due to circleci security feature.

PR #29, opened 2020-08-04
Feature/369: Shorter Import Paths Using Absolute TypeScript Paths

ramanmangla:
In reference to tickets #369 .

This adds absolute import paths for the source code to the `Auth-Service`.
 - Modify Jest config files
 - Modify `tsconfig.json`
 - Modify `.eslintrc.js`
 - Modify `package.json`
 - Add `tsconfig.build.json`
 - Replace all long import paths with the new implementation


jest.config.js
```diff
@@ -16,5 +19,8 @@ module.exports = {
       lines: 90
     }
   },
+  moduleNameMapper: pathsToModuleNameMapper(compilerOptions.paths, {
```
stevenwjy, line 22: You also need this in the `jest.bdd.config.js` to pass the BDD tests

.eslintrc.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   parser: '@typescript-eslint/parser', // Specifies the ESLint parser
   plugins: [
     'cucumber',
-    '@typescript-eslint'
+    '@typescript-eslint',
```
lewisdaly, line 5: Why a trailing comma here, and then not down below? Maybe we should remove trailing commas altogether

ramanmangla: Okay, I'll remove all the trailing commas for consistency

ramanmangla: But actually more lines already have trailing commas. What to do ? @lewisdaly

jest.config.js
```diff
@@ -1,5 +1,10 @@
 'use strict'
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
kleyow, line 3: https://github.com/mojaloop/auth-service/pull/29/files#diff-e4403a877d80de653400d88d85e4801aR59
this rule negates the need for this yes?

package.json
```diff
@@ -13,16 +13,16 @@
   "scripts": {
     "audit:resolve": "SHELL=sh resolve-audit",
```
stevenwjy, line 14: would it be better to put ` --production` here to avoid having an enormous size of `audit-resolve.json` file?

src/lib/requests.ts
```diff
@@ -32,9 +32,11 @@
 /* istanbul ignore file */
 // Testing will be covered in #354
 
-import Logger from '@mojaloop/central-services-logger'
 import { config } from './config'
-import { ThirdpartyRequests, BaseRequestConfigType } from '@mojaloop/sdk-standard-components'
+import Logger, {
```
stevenwjy, line 36: Just checking, `sdk-standard-components` has `Logger` as a default import?

ramanmangla: Yes @stevenwjy . Both `central-services` and `sdk-standard-components` have `Logger`s.

package.json
ramanmangla: @lewisdaly what do you think?

lewisdaly: Yes please. That's what we're doing elsewhere. But feel free to do that in the next pr you make - no need to do it now if your audit check is passing

eoln:
@ramanmangla is `npm run start` and `src/clit.ts` working well in the command line?

PR #31, opened 2020-08-06
Knex instantiation based on DB_ENV

spikerheado1234:
Ensure that the knex instantiation is based of the config DB_ENVIRONMENT variable.


spikerheado1234:
Closing because git history is messed up.

PR #32, opened 2020-08-06
fix: add config dependent knex instance generation

spikerheado1234:
Add config environment dependent Knex instance.


spikerheado1234:
duplicate effort with phasing out one config file.

PR #33, opened 2020-08-06
chore(circleci): add steps to circleci to automate github release

kleyow:


.circleci/config.yml
```diff
@@ -266,23 +269,98 @@ jobs:
...
+          command: |
+            git diff --no-indent-heuristic master~2 HEAD CHANGELOG.md | sed -n '/^+[^+]/ s/^+//p' > /tmp/changes
+            echo 'export RELEASE_CHANGES=`cat /tmp/changes`' >> $BASH_ENV
+      - run:
+          name: check the release changes
+          command: |
+            echo "Changes are: ${RELEASE_CHANGES}"
+      - github-release/create:
+          github-token-variable: ${GITHUB_TOKEN}
+          tag: ${CIRCLE_TAG}
```
lewisdaly, line 326: You won't be able to use ${CIRCLE_TAG} here since you are on the master branch.

.circleci/config.yml
```diff
@@ -266,23 +269,98 @@ jobs:
...
+          name: Generate changelog and bump package version
+          command: npm run release
+      - run:
+          name: Push the release
+          command: git push --follow-tags origin ${CIRCLE_BRANCH}
+      - run:
+          # Note: this is rather imperfect, but will do for now
+          name: Format the changelog into the github release body
+          command: |
+            git diff --no-indent-heuristic master~2 HEAD CHANGELOG.md | sed -n '/^+[^+]/ s/^+//p' > /tmp/changes
```
lewisdaly, line 318: This may need to be `master~1` instead, since you have only just pushed the changes in the previous command. There might be a better way to get this changelog...

.circleci/config.yml
```diff
@@ -266,23 +269,98 @@ jobs:
...
+          command: docker save -o /tmp/docker-image.tar $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:local
       - persist_to_workspace:
           root: /tmp
           paths:
             - ./docker-image.tar
 
+  release:
+    executor: default-docker
+    steps:
+      - checkout
```
lewisdaly, line 292: Maybe move this to after the install dependencies to keep things consistent

.circleci/config.yml
lewisdaly: If you put this one command ahead (before the `git push`), you could just run `git diff CHANGELOG.md | sed ...`

.circleci/config.yml
```diff
@@ -266,23 +269,98 @@ jobs:
...
+          command: |
+            docker tag $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:local $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:latest
+            docker tag $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:local $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:${CIRCLE_TAG}
+      - run:
+          name: Publish Docker image latest tag to Docker Hub
+          command: |
+            echo "Publishing $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:latest"
+            docker push $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:latest
+            echo "Publishing $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:${CIRCLE_TAG}"
+            docker push $DOCKER_ORG/$CIRCLE_PROJECT_REPONAME:${CIRCLE_TAG}
```
lewisdaly, line 362: Maybe after this we should add a fail only slack notification for the `#announcements-ci-cd` channel

.circleci/config.yml
```diff
@@ -428,99 +440,151 @@ workflows:
...
+          filters:
+            tags:
+              only: /.*/
+
       - setup:
           context: org-global
+          filters:
+            tags:
+              only: /.*/
+            branches:
```
lewisdaly, line 454: I don't think we really need these branch ignores here

.circleci/config.yml
kleyow: going to move `checkout` to before dependency install instead.

.circleci/config.yml
kleyow: sounds good.

.circleci/config.yml
kleyow: should I move this back to `publish` then?

.circleci/config.yml
kleyow: done.

.circleci/config.yml
kleyow: is this refering to `setup` only?

.circleci/config.yml
lewisdaly: Nah you don't want to do that. I had it like that earlier, but if git isn't installed, circleci defaults to a bad version of git which can't checkout tags

.circleci/config.yml
lewisdaly: Nah you will just have to use the powers of bash to get the tag, and set it as an environment variable. 

maybe something like:
```
NEW_TAG="v`cat package.json | jq -r .version`"
```

You could also do `git tag | head -1`, but semantic versioning doesn't sort lexicographically, so it might be tough to ensure you are getting the latest

.circleci/config.yml
lewisdaly: I mean for any branch ignore - apart from the `publish` step

.circleci/config.yml
```diff
@@ -65,10 +65,10 @@ jobs:
   setup:
     executor: default-docker
     steps:
+      - checkout
```
eoln: don't do checkout before installing dependencies -> for tagged branch builds (for branches different than master) the circle-ci default git client is failing

.circleci/config.yml
kleyow: done.

.circleci/config.yml
kleyow: good to know. done. added a comment as well.

.circleci/config.yml
kleyow: done.

kleyow:
@lewisdaly one last pass if you can

PR #41, opened 2020-08-18
feature(put consents): signature validation handler

akapur99:
To address ticket mojaloop/mojaloop#340, endpoint to deal with Signature Validation.


test/unit/server/handlers/consents/{ID}.test.ts
```diff
@@ -0,0 +1,325 @@
...
+        expect(mockLoggerPush).not.toHaveBeenCalled()
+      })
+
+    it('should propagate retrieveValidConsent IncorrectChallenge error',
+      async (): Promise<void> => {
+        mockRetrieveValidConsent.mockRejectedValueOnce(new IncorrectChallengeError(consentId))
+
+        await expect(Handler.validateAndUpdateConsent(consentId, credentialRequest, destinationParticipantId)).resolves.toBeUndefined()
+
+        expect(mockLoggerPush).toBeCalledWith(new IncorrectChallengeError(consentId))
```
lewisdaly, line 215: I really don't like validating our tests based on what the logger did - perhaps we should think about how we are throwing and catching errors to we can test this properly.

test/unit/server/handlers/consents/{ID}/generateChallenge.test.ts
```diff
@@ -126,8 +127,8 @@ const completeConsentRevoked: Consent = {
 }
 
 const credential: ConsentCredential = {
-  credentialType: 'FIDO',
-  credentialStatus: 'PENDING',
+  credentialType: 'FIDO' as 'FIDO',
```
lewisdaly, line 130: Why this cast?

package.json
```diff
@@ -14,6 +14,7 @@
     "audit:resolve": "SHELL=sh resolve-audit --production",
     "audit:check": "SHELL=sh check-audit --production",
     "build": "tsc -p ./tsconfig.build.json",
+    "build:openapi": "multi-file-swagger ./src/interface/template.yaml > src/interface/swagger.json",
```
lewisdaly, line 17: Hmm we should probably also update the readme to recommend this - I'm sure there's docs we can borrow from `thirdparty-api-adapter` or `thirdparty-scheme-adapter`

lewisdaly (changes requested):

package.json
akapur99: can't seem to find docs like that there but will update the readme

test/unit/domain/consents/{ID}.test.ts
```diff
@@ -0,0 +1,368 @@
...
+    payload: publicKey
+  }
+}
+
+describe('server/domain/consents/{ID}', (): void => {
+  beforeAll((): void => {
+    mockLoggerError.mockReturnValue(null)
+    mockLoggerPush.mockReturnValue(null)
+
+    mockConsentDbRetrieve.mockResolvedValue(retrievedConsent)
```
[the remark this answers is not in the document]
akapur99: so i guess the logic here is to declare best case mocks here - and then for each test, just change mocking for needed functions for that test

which prevents us from having to redeclare every mock for each test - i thought this might improve readability? Though I do see your point here

src/domain/consents/{ID}.ts
```diff
@@ -0,0 +1,105 @@
...
+  const scopes: Scope[] = await scopeDB.retrieveAll(consent.id)
+  const externalScopes: ExternalScope[] = convertScopesToExternal(scopes)
+
+  const consentBody: PutConsentsRequest = {
+    requestId: consent.id,
+    scopes: externalScopes,
+    initiatorId: consent.initiatorId as string,
+    participantId: consent.participantId as string,
+    credential: {
+      id: consent.credentialId as string,
```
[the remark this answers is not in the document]
akapur99: in this function - we can probably be defensive and add checks 
but in the whole flow for this endpoint - JOI should already check for this in the incoming request and even if not there - the error will get caught in the handler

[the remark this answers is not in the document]
lewisdaly: Certainly, but I think our type definitions aren't really well defined then, since we shouldn't have to cast here

test/unit/server/handlers/consents/{ID}.test.ts
```diff
@@ -0,0 +1,325 @@
...
+const mockBuildConsentRequestBody = jest.spyOn(Domain, 'buildConsentRequestBody')
+const mockLoggerPush = jest.spyOn(Logger, 'push')
+const mockLoggerError = jest.spyOn(Logger, 'error')
+const mockVerifySignature = jest.spyOn(Signature, 'verifySignature')
+
+/*
+ * Mock Request Resources
+ */
+// @ts-ignore
+const request: Request = {
```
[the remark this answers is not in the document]
akapur99: makes sense - we can make this a separate PR

test/unit/server/handlers/consents/{ID}/generateChallenge.test.ts
akapur99: We haven't declared a credentialTypeEnum - but I think after what we've done here - when we're updating type definitions we should refactor this there ?

test/unit/server/handlers/consents/{ID}.test.ts
akapur99: Ah i see - i think either works - as both are place holders for eventually making a `PUT .../error` call

test/unit/server/handlers/consents/{ID}/generateChallenge.test.ts
lewisdaly: Yeah but I don't understand why you would need to cast a string to itself in the first place

src/model/consent/consent.ts
```diff
@@ -65,9 +65,16 @@ export interface Consent {
 /*
  * Interface for Consent Credential resource type
  */
+
+export enum CredentialStatusEnum {
+  VERIFIED = 'VERIFIED',
+  PENDING = 'PENDING',
+}
+
 export interface ConsentCredential {
```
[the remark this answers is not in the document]
akapur99: will be handled in sep PR

src/interface/definitions.yaml
```diff
@@ -108,6 +135,40 @@ Credential:
...
+      type: string
+      description: Base64 encoded bytes - The public key of the Public/Private keypair
+  required:
+    - id
+    - type
+    - status
+    - challenge
+    - payload
+
  # TODO: update me to be an empty body!
```
lewisdaly, line 172: Remove todo

test/unit/domain/consents/{ID}.test.ts
```diff
@@ -0,0 +1,369 @@
...
+    challenge: {
+      signature,
+      payload: challenge
+    },
+    payload: publicKey,
+    id: requestCredentialId
+  }
+} = request.payload as UpdateCredentialRequest
+
+/* TODO, fill out later. */
```
lewisdaly, line 170: Remove todo

test/unit/domain/consents/{ID}.test.ts
```diff
@@ -0,0 +1,369 @@
...
+
+        expect(mockConsentDbUpdate).not.toBeCalled()
+
+        // Reset payload
+        credentialVerified.credentialPayload = 'string_representing_credential_payload'
+      })
+
+    it('should throw error if credential payload empty string',
+      async (): Promise<void> => {
+        // Make Credential Payload undefined
```
lewisdaly, line 320: change comment

test/unit/domain/consents/{ID}.test.ts
```diff
@@ -0,0 +1,369 @@
...
+      async (): Promise<void> => {
+        mockConsentDbUpdate.mockRejectedValueOnce(new Error('ConsentDB Error'))
+        await expect(updateConsentCredential(retrievedConsent, credentialVerified))
+          .rejects
+          .toThrowError('ConsentDB Error')
+
+        expect(mockConsentDbUpdate).toBeCalledWith(updatedConsent)
+      })
+
+    it('should throw error if credential payload undefined',
```
lewisdaly, line 303: change test name to `null`

lewisdaly (approved):

PR #44, opened 2020-09-02
Use api snippets for auth-service interfaces

kkzeng:
- Remove old files
- Replace all references with api-snippet references
- Change buildopen:api script to resolve api-snippet references and validate new api.yaml


lewisdaly (commented):
Awesome! Thanks Kenneth. 

Btw since we changed the file name from `swagger.json`, we should update the [documentation here](https://github.com/mojaloop/auth-service#updating-the-openapi-swagger-spec)

kleyow:
looks good to me

src/server/handlers/metrics.ts
```diff
@@ -26,7 +26,11 @@
 import Metrics from '@mojaloop/central-services-metrics'
 import { Request, ResponseToolkit, ResponseObject } from '@hapi/hapi'
 
-export function get (_: Request, h: ResponseToolkit): ResponseObject {
+function get (_: Request, h: ResponseToolkit): ResponseObject {
```
eoln, line 29: I would keep export here to have the same consistency wich we keep in other handlers

src/shared/histogram.ts
```diff
@@ -0,0 +1,34 @@
...
+export type THandlerFunc = (request: Request, h: ResponseToolkit) => Promise<ResponseObject>
+
+/**
+ * @function wrapWithHistogram
+ * @description Wraps a handler function with a histogram of the given name
+ * @param {THandlerFunc} handler The handler function to be wrapped
+ * @param {[string, string, Array<string>]} histogramParams The params of the histogram
+ */
+function wrapWithHistogram (handler: THandlerFunc, histogramParams: [string, string, string[]]): THandlerFunc {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
eoln, line 14: where do we have `any` here? drop the linting exception

eoln (changes requested):
really small things to fix

kkzeng:
Woohoo, finally CI is green :D Time to address comments

src/shared/histogram.ts
kkzeng: Sure, I can remove it. Not sure why it's there either.
I was wondering if I need the histogram for auth-service?
I removed the histogram in my latest commit but I can add it back

src/server/handlers/metrics.ts
kkzeng: yup agreed

src/server/handlers/metrics.ts
```diff
@@ -26,7 +26,11 @@
 import Metrics from '@mojaloop/central-services-metrics'
 import { Request, ResponseToolkit, ResponseObject } from '@hapi/hapi'
 
-export function get (_: Request, h: ResponseToolkit): ResponseObject {
+function get (_context: any, _: Request, h: ResponseToolkit): ResponseObject {
```
eoln, line 29: use here Context interface you defined in`~/server/plugins`

eoln (changes requested):

src/server/handlers/metrics.ts
kkzeng: yes, I think I didn't save in VSCode, I fixed it now. thanks for catching this

lewisdaly (approved):
Looks good! Thanks @kkzeng

eoln (approved):
lgtm

PR #45, opened 2020-09-12
feat(test): api route testing

kkzeng:


test/unit/index.test.ts
```diff
@@ -26,74 +26,179 @@
...
+})
+
+describe('api routes', (): void => {
+  let server: Server
+
+  beforeAll(async (): Promise<Server> => {
+    mockLoggerPush.mockReturnValue(null)
+    mockLoggerError.mockReturnValue(null)
+    server = await index.server.run(Config)
+    return server
```
lewisdaly, line 62: No need to return the server here

test/unit/index.test.ts
```diff
@@ -26,74 +26,179 @@
...
+    const request = {
+      method: 'POST',
+      url: '/consents',
+      headers: Headers,
+      payload: MockConsentData.payload
+    }
+
+    const response = await server.inject(request)
+    expect(response.statusCode).toBe(202)
+    expect(response.result).toBeDefined()
```
lewisdaly, line 139: We should also check that `mockPostConsent` was called once, and also verify that what it was called with matches what we expect.

lewisdaly: (same goes for all the others)

src/server/handlers/consents/{ID}.ts
```diff
@@ -119,7 +120,7 @@ export async function put (_context: Context, request: Request, h: ResponseToolk
   const destinationParticipantId = request.headers[Enum.Http.Headers.FSPIOP.SOURCE]
 
   // Note: not awaiting promise here
-  validateAndUpdateConsent(id, updateConsentRequest, destinationParticipantId)
+  thisModule.validateAndUpdateConsent(id, updateConsentRequest, destinationParticipantId)
```
lewisdaly, line 123: I'm not sure I understand the need for this here - is that just to get the mocks to work?

lewisdaly (dismissed):

src/server/handlers/consents/{ID}.ts
kkzeng: Yes, I tested without doing this and the original function is still called.
This was what I referenced: https://stackoverflow.com/questions/45111198/how-to-mock-functions-in-the-same-module-using-jest/45288360

test/unit/index.test.ts
```diff
@@ -131,7 +131,15 @@ describe('api routes', (): void => {
...
     }
 
+    const expectedArgs = expect.objectContaining({
+      path: '/consents',
+      method: 'post',
+      payload: MockConsentData.payload
+    })
+
     const response = await server.inject(request)
+    expect(mockCreateConsent).toHaveBeenCalledTimes(1)
```
eoln, line 140: we are testing here only that our handler has been called, but the handler itself is mocked, so the code of handler isn't tested here - do we have handlers tested elsewhere?

eoln: I don't see that `CreateConsent` https://github.com/mojaloop/auth-service/blob/master/src/server/handlers/index.ts#L40
is implemented anywhere...

eoln (changes requested):

test/unit/index.test.ts
kkzeng: > we are testing here only that our handler has been called, but the handler itself is mocked, so the code of handler isn't tested here - do we have handlers tested elsewhere?

Yes, the handlers are here: ` https://github.com/mojaloop/auth-service/tree/master/test/unit/server/handlers`

test/unit/data/data.ts
```diff
@@ -83,7 +83,7 @@ export const requestWithPayloadCredentialAndScope: Request = {
 // @ts-ignore
```
eoln, line 83: do we need this @ts-ignore ?

eoln (approved):

test/unit/data/data.ts
kkzeng: @eoln it is required because the Request interface has many more fields. So in order to create an object with only the relevant fields, they used ts-ignore I guess.

PR #46, opened 2020-09-16
Remove istanbul-ignore from handlers

kkzeng:
Make sure handlers are also tested for coverage

### DIFF
--- a/jest.bdd.config.js
+++ b/jest.bdd.config.js
@@ -15,7 +15,7 @@ module.exports = {
     global: {
       statements: 20,
       functions: 20,
-      branches: 20,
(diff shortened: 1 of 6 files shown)


lewisdaly:
I would just [lower this](https://github.com/mojaloop/auth-service/blob/master/jest.bdd.config.js#L14) so that test-bdd passes.

kkzeng:
> I would just [lower this](https://github.com/mojaloop/auth-service/blob/master/jest.bdd.config.js#L14) so that test-bdd passes.

Ok, sounds good 👍

lewisdaly (approved):

PR #47, opened 2020-09-17
Master Convict config 

kkzeng:


scripts/wait4.config.js
```diff
@@ -12,16 +12,22 @@ module.exports = {
   // services definitions
   services: [
     {
-      name: 'my-sql',
+      name: 'containers',
```
eoln, line 15:
```suggestion
      name: 'cicd-integration-tests',
```

seeds/01_consent.ts
```diff
@@ -95,6 +96,7 @@ const consents = [
   }
 ]
 
-export function seed (knex: Knex): Knex.QueryBuilder<number[]> {
-  return knex('Consent').insert(consents)
+export function seed (knex: Knex): Promise<Knex.QueryBuilder<number[]>> {
+  return knex('Consent').del()
```
eoln, line 100: do we really want to delete all entries? please add a comment on why.

seeds/02_scope.ts
```diff
@@ -49,6 +52,7 @@ const scopes = [
   }
 ]
 
-export function seed (knex: Knex): Knex.QueryBuilder<number[]> {
-  return knex('Scope').insert(scopes)
+export function seed (knex: Knex): Promise<Knex.QueryBuilder<number[]>> {
+  return knex(tableName).del()
```
eoln, line 56: let keep consistency and do not keep the table name in the external variable but use the string value instead `Scope` as you are doing it in the line after

src/shared/config.ts
```diff
@@ -24,60 +24,145 @@
...
+    SHOW_HIDDEN: boolean;
+    COLOR: boolean;
   };
 }
 
-const RC = parse(rc('AS', Config)) as ServiceConfig
+const ConvictConfig = Convict<ServiceConfig>({
+  SHOULD_USE_IN_MEMORY_DB: {
+    doc: 'Should we use the in memory database',
+    format: ['true', 'false'],
```
eoln, line 92:
```suggestion
    format: 'Boolean',
```

src/shared/config.ts
```diff
@@ -24,60 +24,145 @@
...
+    default: 'false',
```
eoln, line 93:
```suggestion
    default: false,
```

src/shared/config.ts
```diff
@@ -24,60 +24,145 @@
...
+let DBConfig: DatabaseConfig = DbConfig.development
+
+if (shouldUseInMemoryDB) {
+  DBConfig = DbConfig.test
+}
+
+// Load and validate general config
+ConvictConfig.loadFile(`${__dirname}/../../config/config.json`)
+ConvictConfig.validate({ allowed: 'strict' })
+
```
eoln, line 148: here should be database config section, not before loading config from json

src/shared/config.ts
```diff
@@ -24,60 +24,145 @@
...
+    },
+    COLOR: {
+      doc: 'Show colors in output',
+      format: 'Boolean',
+      default: true
+    }
+  }
+})
+
+// Load and validate database config
```
eoln, line 137: database config section should be after loading and validating config from json

src/shared/config.ts
```diff
@@ -24,60 +24,145 @@
...
+  SHOULD_USE_IN_MEMORY_DB: string;
+  INSPECT: {
+    DEPTH: number;
+    SHOW_HIDDEN: boolean;
+    COLOR: boolean;
   };
 }
 
-const RC = parse(rc('AS', Config)) as ServiceConfig
+const ConvictConfig = Convict<ServiceConfig>({
```
eoln, line 89: there is a need to add ENV variable like it has been done here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/shared/config.ts#L121

src/shared/config.ts
```diff
@@ -24,60 +24,145 @@
...
+    loadExtensions: string[];
   };
+}
 
-  // inspect.ts
-  INSPECT?: {
-    DEPTH?: number;
-    SHOW_HIDDEN?: boolean;
-    COLOR?: boolean;
+interface ServiceConfig {
```
eoln, line 76: add ENV prop as here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/shared/config.ts#L53

src/shared/config.ts
```diff
@@ -24,60 +24,145 @@
...
+// Load and validate database config
+const shouldUseInMemoryDB = ConvictConfig.get('SHOULD_USE_IN_MEMORY_DB') === 'true'
+let DBConfig: DatabaseConfig = DbConfig.development
+
+if (shouldUseInMemoryDB) {
+  DBConfig = DbConfig.test
+}
+
+// Load and validate general config
+ConvictConfig.loadFile(`${__dirname}/../../config/config.json`)
```
eoln, line 146: there is a need for separate config files for every environment as it is done here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/shared/config.ts#L284-L286

config/config.json
```diff
@@ -0,0 +1,11 @@
+{
```
eoln, line 1: there is a need for separate config files pere environment: https://github.com/mojaloop/thirdparty-scheme-adapter/tree/master/config
you should also specify here ENV & SHOULD_USE_IN_MEMORY_DB missed props

config/knexfile.ts
```diff
@@ -1,25 +1,79 @@
...
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Kenneth Zeng <kkzeng@google.com>
+ --------------
+ ******/
+
+import { DatabaseConfig } from '../src/shared/config'
 import path from 'path'
+require('ts-node/register')
```
eoln, line 32: what for do we need `ts-node/register` ?

eoln (changes requested):
- please use Convict for knex config
- create knex config files for every environment
- use Convict validation
- use ENV variable for selecting environment

seeds/01_consent.ts
kkzeng: from [the knex documentation](http://knexjs.org/#Seeds-CLI): "You should design your seed files to reset tables as needed before inserting data."

Our seed files were not resetting the table as needed so when we would seed a table more than once, there would be duplicate entry errors etc.

config/knexfile.ts
kkzeng: removed it

src/shared/config.ts
kkzeng: done :)

seeds/01_consent.ts
kkzeng: So we want to clear all the entries before populating them again with the entries defined in the seed file so that we are in a good state.

seeds/02_scope.ts
kkzeng: agreed

src/shared/config.ts
kkzeng: Sorry, not sure what you mean by this comment @eoln

config/knexfile.ts
```diff
@@ -1,58 +1,197 @@
...
+    loadExtensions: {
+      doc: 'Array of extensions to load',
+      format: 'Array',
+      default: ['.ts']
     }
   }
-}
+})
+
+const env = process.env.NODE_ENV ?? 'production'
```
kkzeng, line 184: @eoln I was unable to import the env from the shared Config file. Should I be doing this or should I add the environment variable to the schema here as well?

config/knexfile.ts
```diff
@@ -1,58 +1,197 @@
...
+const dbConfigFile = `${__dirname}/${env}_db.json`
+ConvictDatabaseConfig.loadFile(dbConfigFile)
+ConvictDatabaseConfig.validate({ allowed: 'strict' })
+
+const Config: DatabaseConfig = ConvictDatabaseConfig.getProperties()
+
+// Inject directory paths here
```
kkzeng, line 191: @eoln I removed the directory specification from config files themselves because the path resolution causes problems without using __dirname and resolving the absolute paths and I can't do that in the config file itself since its json and not script.

config/knexfile.ts
eoln: @kkzeng please add the ENV variable to the schema like here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/shared/config.ts#L121-L126

eoln: in definition of format for ENV you should specify the list of allowed values for your environments

config/custom-convict-formats.ts
```diff
@@ -0,0 +1,32 @@
+import { DbConnection } from './knexfile'
```
eoln, line 1: I like how you handle this small challenge 👍

config/default.json
```diff
@@ -1,32 +1,32 @@
...
-    "DISABLED": false,
-    "RUN_DATA_MIGRATIONS": true
-  },
-  "INSPECT": {
-    "DEPTH": 4,
-    "SHOW_HIDDEN": false,
-    "COLOR": true
-  }
-}
+    "PORT": 4004,
```
eoln, line 2: do we still need default.json file?

src/shared/config.ts
eoln: you already did it in a correct way

config/knexfile.ts
kkzeng: Right, I wasn't sure if I should duplicate the variable again since in the `~/src/shared/config.ts` we already have the env. So we can add it again to DatabaseConfig schema?

eoln (approved):
lgtm, no need for config/default.json file anymore

config/knexfile.ts
eoln: no need for ENV in DatabaseConfig -> you are resolving which *_db.ts file to load using the ENV prop defined in ~/src/shared/config

eoln:
unit tests are failing @kkzeng

eoln (changes requested):
unit tests are failing

config/knexfile.ts
kkzeng: So that's what I was saying before - when I import env from ~/src/shared/config, it is `undefined` for some reason. I tried exporting the env variable itself.

So in ~/src/shared/config, I had something like:
`export const env = ConvictConfig.get('ENV')`

config/default.json
kkzeng: no need

eoln (approved):
lgtm

config/knexfile.ts
eoln: @kkzeng take a look here: https://github.com/mojaloop/auth-service/blob/master/tsconfig.json#L44
the `~/` is shorthand so you should import from `~/shared/config` and skip `src` part

PR #37, opened 2020-08-12
[WIP] Feature/388: Setup a master Convict project Config 

ramanmangla:
This is in reference to ticket #388 .

- Replace RC with Convict
- Merge multiple configs into a master config (`src/shared/config.ts`)
- Define Convict Config definition for validation
- Auto-load config properties for dev, test and prod environments using `NODE_ENV`
- Auto-migrate DB after connection setup
- Refactor files to work with new config

PR #48, opened 2020-09-24
Integration tests: handlers and domain

kkzeng:
- Integration tests for all domains
- Integration tests for all handlers (202 Accepted response)
- Remove coverage check


Dockerfile
```diff
@@ -18,6 +18,8 @@ RUN adduser -D as-user
 USER as-user
 COPY --chown=as-user --from=builder /opt/auth-service .
 # cleanup
-RUN npm prune --production
+
+# disable until problems with TS->JS transpilation solved
+# RUN npm prune --production
```
eoln, line 23: or move ts-node to prod dependencies...

config/custom-convict-formats.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (typeof val === 'string' || val instanceof String) return true;
+
+        //  Object check
+        if (typeof val === 'object' && val !== null) {
+            const connection = val as DbConnection;
+
+            // Check that object is DbConnection and has DbConnection fields - 
+            // i.e. PG and MySQL use a DbConnection object to configure
+            // Verify that all fields are filled AND they are the right format - presence + format check
+            if (connection.host == null || typeof connection.host !== 'string') {
```
eoln, line 16:
```suggestion
            if (!(typeof connection?.host === 'string')) {
```
let apply this compact ?. style also below

scripts/wait4.js
```diff
@@ -106,21 +106,14 @@ async function wrapWithRetries (method, waitJob, retries, waitTimeMs) {
...
-  const namespace = (waitJob.rc && waitJob.rc.namespace) || 'CLEDG'
-  const configPath = (waitJob.rc && waitJob.rc.configPath) || '../config/default.json'
-
-  // require rc to deliver config
-  try {
-    return require('rc')(namespace, require(configPath))
-  } catch (err) {
-    return waitJob.rc || {}
-  }
+  const configPath = '../config/integration_db.json'
```
eoln, line 115: maybe we should resolve config path using NODE_ENV variable, so we will be able to run integration tests with different configurations for knex...

src/shared/config.ts
```diff
@@ -24,60 +24,91 @@
  - Name Surname <name.surname@gatesfoundation.com>
 
  - Paweł Marzec <pawel.marzec@modusbox.com>
- - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+ - Raman Mangla <ramanmangla@google.com>
```
eoln, line 27: why you drop @akapur99 ?

test/integration/domain/consents.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    await closeKnexConnection()
+  })
+
+  it('Should resolve successfully', async (): Promise<void> => {
+    mockConvertExternalToScope.mockReturnValueOnce(scopes)
+    await expect(createAndStoreConsent(requestWithPayloadScopes))
+      .resolves
+      .toBe(undefined)
+  })
+})
```
eoln, line 50: you should check how `mockConvertExternalToScope` has been called

eoln: why are you mocking some code in integration tests?

test/integration/domain/consents/generateChallenge.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      const credentialUpdate: ConsentCredential = {
+        credentialType: 'FIDO',
+        credentialStatus: CredentialStatusEnum.PENDING,
+        credentialChallenge: 'string_representing_credential_challenge',
+        credentialPayload: null
+      }
+
+      const updatedConsent: Consent = await updateConsentCredential(consentMissingCred, credentialUpdate)
+
+      const expectedConsent: Consent = Object.assign({}, consentMissingCred)
```
eoln, line 76:
```suggestion
      const expectedConsent: Consent = { ...consentMissingCred }
```

eoln (changes requested):
some code style comments

if we are testing the async API POST method, shouldn't we verify that corresponding PUT method has been called correctly?

src/shared/config.ts
kkzeng: Not me who did this :( it was @ramanmangla

kkzeng:
> some code style comments
> 
> if we are testing the async API POST method, shouldn't we verify that corresponding PUT method has been called correctly?

That's a good point - I think I saw that the test that you are mentioning already existed in the handler unit tests - I will move them over to the integration tests or write this if I am mistaken about the unit tests already having that

eoln:
Unit tests are using mocked endpoints. The clue of integration tests is to test how our code is integrated with these endpoints.

kkzeng:
> Unit tests are using mocked endpoints. The clue of integration tests is to test how our code is integrated with these endpoints.

Right, so you mean we would need to use the Mojaloop testing toolkit right?
In standup, Lewis said that we should move that to a different ticket and put it off for now?

config/custom-convict-formats.ts
kkzeng: @eoln now that I think about it, do we even need the optional because here we are saying that connection could be null when in reality, we assert above that connection is not null

kkzeng: Maybe we can just use `connection.host !== 'string'`

src/shared/config.ts
kkzeng: You can see here: https://github.com/mojaloop/auth-service/pull/37/files
I will add him back :)

scripts/wait4.js
kkzeng: Makes sense but I will have to use `process.env.NODE_ENV` because I cannot import typescript files here - so I cannot retrieve it from Convict unless we make another javascript version

test/integration/domain/consents.test.ts
kkzeng: I am also not too sure why I mocked it, I removed the mocks now :)

Dockerfile
kkzeng: @eoln Turns out that this line also breaks integration tests because axios is also a devDependency. 
Moving axios to prod dependencies doesn't make sense since it is only used for testing.

.circleci/config.yml
```diff
@@ -181,8 +181,8 @@ jobs:
           command: |
             echo ${NVM_DIR}
             [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-            nvm install v10.15.3
-            nvm alias default v10.15.3
+            nvm install v12.18.4
```
eoln, line 184: good you update it, but I think it should be also in sync with Dockerfile: https://github.com/mojaloop/auth-service/blob/master/Dockerfile#L1

eoln (approved):

.circleci/config.yml
kkzeng: Ok I will update it to be `12.16.1` to be consistent with the docker image

PR #52, opened 2020-10-05
JWS, TLS Signing and Endpoint Configuration

kkzeng:


src/shared/config.ts
```diff
@@ -44,7 +45,44 @@ interface ServiceConfig {
...
+    ALS_ENDPOINT: string;
+    QUOTES_ENDPOINT: string;
+    TRANSFERS_ENDPOINT: string;
+    BULK_TRANSFERS_ENDPOINT: string;
+    JWS_SIGN: boolean;
+    JWS_SIGNING_KEY: PathLike | Buffer;
+    WSO2_AUTH: {
+      staticToken: string;
+      tokenEndpoint: string;
+      clientKey: string;
```
kkzeng, line 60: I noticed that `thirdparty-scheme-adapter` also added these fields (after Kevin's PR which I referenced):

```
DFSP_BACKEND_URI: {
      doc: 'host address of DFSP\'s ',
      format: '*',
      default: 'localhost:9000'
    },
    DFSP_BACKEND_SIGN_AUTHORIZATION_PATH: {
      doc: 'path use by BackendRequests.signAuthorizationRequest',
      format: '*',
      default: 'signchallenge'
    },
    DFSP_BACKEND_HTTP_SCHEME: {
      doc: 'Http scheme ',
      format: ['http', 'https'],
      default: 'http'
    },
```

@lewisdaly Are these also needed? Also I removed DFSP_ID since we already specify DFSP_ID through PARTICIPANT_ID, is that correct

config/development.json
```diff
@@ -6,6 +6,31 @@
...
+      "clientKey": "test-client-key",
+      "clientSecret": "test-client-secret",
+      "refreshSeconds": 3600
+    },
+    "TLS": {
+      "mutualTLS": {
+        "enabled": false
+      },
+      "creds": {
+        "ca": "./secrets/dfsp_a_client.crt",
```
lewisdaly, line 30: You can remove the `dfsp_a_` prefix, and rename the files accordingly

src/shared/config.ts
lewisdaly: Nah we don't need those fields, and yep, `PARTICIPANT_ID` is correct.

src/shared/config.ts
```diff
@@ -92,6 +130,34 @@ const ConvictConfig = Convict<ServiceConfig>({
       format: 'Boolean',
       default: true
     }
+  },
+  // TODO: Add documentation
```
lewisdaly, line 134: TODO

src/shared/config.ts
```diff
@@ -92,6 +130,34 @@ const ConvictConfig = Convict<ServiceConfig>({
...
+    JWS_SIGN: false,
+    JWS_SIGNING_KEY: '',
+    WSO2_AUTH: {
+      staticToken: '',
+      tokenEndpoint: '',
+      clientKey: '',
+      clientSecret: '',
+      refreshSeconds: 60
+    },
+    // TODO: Investigate proper key setup
```
lewisdaly, line 150: TODO

kkzeng: This TODO was carried over from the `thirdparty-scheme-adapter` not sure how to resolve this

src/shared/config.ts
kkzeng: I need your help to figure out what the config params mean

eoln (approved):
lgtm

src/lib/requests.ts
```diff
@@ -43,19 +43,16 @@ import {
...
-        enabled: false
-      }
-    }
-  }
-
+  peerEndpoint: Config.SHARED.PEER_ENDPOINT,
+  alsEndpoint: Config.SHARED.ALS_ENDPOINT,
+  quotesEndpoint: Config.SHARED.QUOTES_ENDPOINT,
+  transfersEndpoint: Config.SHARED.TRANSFERS_ENDPOINT,
+  bulkTransfersEndpoint: Config.SHARED.BULK_TRANSFERS_ENDPOINT,
```
eoln, line 52: these params are used here: https://github.com/mojaloop/sdk-standard-components/blob/master/src/lib/requests/baseRequests.js#L125-L135

so you need to add all `thirdpartyRequestsEndpoint` 
take a look also here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/server/plugins/state.ts#L83-L111

src/shared/config.ts
eoln: drop these TODO comments

eoln (changes requested):
there is a need to have all endpoints required in sdk-standard-components

eoln (approved):
lgtm

PR #61, opened 2021-03-22
feat: use api-snippets(thirdparty-api paths) in place of local defs

sridharvoruganti:
this is related to story #[2089](https://github.com/mojaloop/project/issues/2089)

@lewisdaly updated api definitions and modified code to support new payload.


kleyow (approved):

migrations/20200624121732_consent.ts
```diff
@@ -32,7 +32,7 @@ export async function up (knex: Knex): Promise<void | Knex.SchemaBuilder> {
         return knex.schema.createTable('Consent',
           (t: Knex.CreateTableBuilder): void => {
             // TODO: Confirm string length for status and revoked at
-            t.string('id', 32).primary().notNullable()
+            t.string('id', 36).primary().notNullable()
```
lewisdaly, line 35: We technically should never modify a migration, but considering this isn't live anywhere yet it should be ok.

test/unit/server/handlers/consents.test.ts
```diff
@@ -68,7 +68,7 @@ describe('server/handlers/consents', (): void => {
       expect(setImmediate).toHaveBeenCalled()
     })
 
-  it('Should return 400 code due to invalid request',
+  xit('Should return 400 code due to invalid request',
```
lewisdaly, line 71: Is this a typo?

src/server/handlers/consents.ts
```diff
@@ -48,18 +49,19 @@ export async function post (
   _context: Context,
   request: Request,
   h: ResponseToolkit): Promise<ResponseObject> {
+  // TODO: is this validation required? participantId removed from payload
   // Validate request
-  if (!isPostConsentRequestValid(request)) {
-    return h.response().code(Enum.Http.ReturnCodes.BADREQUEST.CODE)
-  }
+  // if (!isPostConsentRequestValid(request)) {
```
lewisdaly, line 54: Will this be done by the Hapi server? Maybe we need a test case to make sure (or maybe we already have one?)

lewisdaly (changes requested):
A couple minor nitpicks

src/interface/api-template.yaml
```diff
@@ -8,405 +8,17 @@ info:
...
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/405'
-        406:
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/406'
-        501:
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/501'
-        503:
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/503'
-
+    $ref: '../../node_modules/@mojaloop/api-snippets/thirdparty/openapi3/paths/consents_ID_generateChallenge.yaml'
   /consents/{ID}/revoke:
```
lewisdaly, line 17: We can remove `/revoke`: https://github.com/mojaloop/pisp/blob/master/docs/linking/README.md#21-unlinking-without-a-hub-hosted-auth-service

src/interface/api-template.yaml
```diff
@@ -8,405 +8,17 @@ info:
...
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/405'
-        406:
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/406'
-        501:
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/501'
-        503:
-          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/503'
-
+    $ref: '../../node_modules/@mojaloop/api-snippets/thirdparty/openapi3/paths/consents_ID.yaml'
   /consents/{ID}/generateChallenge:
```
lewisdaly, line 15: We can remove this resource. The challenge is now derived by the PISP. See: https://github.com/mojaloop/pisp/blob/master/docs/linking/README.md#161-deriving-the-challenge

src/interface/api-template.yaml
lewisdaly: We do need to add a `DELETE /consents/{ID}` resource now

migrations/20200624121732_consent.ts
sridharvoruganti: Ok, but as per API definition consentId type is CorrelationId that is  always 36 characters long

test/unit/server/handlers/consents.test.ts
sridharvoruganti: I will add comment that test is not required now, as we removed participant ID from body.
That validates whether request is valid by comparing if source header matches participant ID

src/server/handlers/consents.ts
sridharvoruganti: I will add comment that test is not required now, as we removed participant ID from body.
That validates whether request is valid by comparing if source header matches participant ID

migrations/20200624121732_consent.ts
eoln: I agree with @sridharvoruganti  -> we should add and keep `id` constraint

eoln: but I also agree with @lewisdaly that we shouldn't touch migration code -> but how to properly handle if migration code contains such a bug like here @lewisdaly ?

src/__mocks__/server.ts
```diff
@@ -26,6 +26,7 @@
  ******/
 
 // mock for server.run
+// eslint-disable-next-line @typescript-eslint/ban-types
```
eoln, line 29: better would be to introduce a new interface to be used here as mock function return type instead of `object`.

src/server/handlers/consents.ts
eoln: openAPI/swagger allows specifying per property constraints. More complex constraints spread over multiple properties/headers must be validated by custom code

src/lib/db.ts
```diff
@@ -32,6 +32,7 @@ import Config from '../shared/config'
 import ConsentDB from '../model/consent'
 import ScopeDB from '../model/scope'
 
+// eslint-disable-next-line @typescript-eslint/ban-types
 const Db: Knex = Knex(Config.DATABASE as object)
```
eoln, line 36: here shouldn't typecasting be needed - if it is then Config.DATABASE is incompatible with Knex constructor signature

src/lib/scopes.ts
```diff
@@ -60,12 +60,14 @@ export function convertScopesToExternal (
     const accountId: string = scope.accountId
 
     if (!(accountId in scopeDictionary)) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
eoln, line 63: let not silence complaints, better is to drop @ts-ignore and apply properly types

src/server/create.ts
```diff
@@ -24,7 +24,7 @@
  ******/
 
 // workaround for lack of typescript types for mojaloop dependencies
-// eslint-disable-next-line @typescript-eslint/no-triple-slash-reference
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
```
eoln, line 27: if ambient.d.ts file is in root directory we should be able to drop `<reference path`

src/server/handlers/consents.ts
eoln: maybe we should drop `isPostConsentRequestValid` from domain/consents.ts and drop this commented code.

src/server/handlers/consents.ts
```diff
@@ -48,18 +49,19 @@ export async function post (
...
+  // }
   // Asynchronously deals with creation and storing of consents and scope
   setImmediate(async (): Promise<void> => {
     try {
       await createAndStoreConsent(request)
     } catch (error) {
       logger.push(error).error('Error: Unable to create/store consent')
-      if(isMojaloopError(error)) {
-        const consentId = (request.payload as PostConsentPayload).id
+      if (isMojaloopError(error)) {
```
eoln, line 63: I think that in case of `any error` and not only when we have `isMojaloopError` we should notify the source via `putConsentError`

eoln: isMojaloopError: https://github.com/mojaloop/auth-service/blob/master/src/domain/errors.ts#L31

src/shared/config.ts
```diff
@@ -158,7 +159,7 @@ const ConvictConfig = Convict<ServiceConfig>({
 // Load and validate general config based on environment variable
 const env = ConvictConfig.get('ENV')
 
-ConvictConfig.loadFile(`${__dirname}/../../config/${env}.json`)
+ConvictConfig.loadFile(path.join(__dirname, `/../../config/${env}.json`))
```
eoln, line 162: @lewisdaly Maybe we should drop using multi-stage config support as we did for thirdparty-api-adapter ? If yes then let do it in a separate story.

test/data/data.ts
```diff
@@ -50,10 +50,10 @@ export const requestWithPayloadCredentialAndScope: Request = {
     fspiopdestination: 'dfsp-3333-2123'
```
eoln, line 50: is `fspiopdestination` in schema?

src/domain/consents.ts
```diff
@@ -42,6 +42,9 @@ import { logger } from '~/shared/logger'
 import { Enum } from '@mojaloop/central-services-shared'
 import { ExternalScope, convertExternalToScope } from '../lib/scopes'
 import { DatabaseError } from './errors'
+import {
+  thirdparty as tpAPI
+} from '@mojaloop/api-snippets'
 
 export interface PostConsentPayload {
```
eoln, line 49: should we update this interface? initiatorId & participantId are coming via headers not payload

test/data/data.ts
```diff
@@ -14,22 +14,22 @@ export const request: Request = {
...
   params: {
-    ID: '1234'
+    ID: 'b51ec534-ee48-4575-b6a9-ead2955b8069'
   },
   payload: {
-    id: '1234',
-    participantId: 'dfsp-3333-2123',
+    consentId: 'b51ec534-ee48-4575-b6a9-ead2955b8069',
+    consentRequestId: 'dfsp-3333-2123',
     initiatorId: 'pisp-2342-2233',
```
eoln, line 33: `initiatorId` should come via header

test/data/data.ts
```diff
@@ -50,10 +50,10 @@ export const requestWithPayloadCredentialAndScope: Request = {
...
   },
   params: {
-    ID: '1234'
+    ID: 'b51ec534-ee48-4575-b6a9-ead2955b8069'
   },
   payload: {
-    id: '1234',
+    id: 'b51ec534-ee48-4575-b6a9-ead2955b8069',
     requestId: '475234',
     initiatorId: 'pispa',
```
eoln, line 58: InitatorId & participantId should come via headers

test/data/data.ts
```diff
@@ -104,29 +104,29 @@ export const h: ResponseToolkit = {
...
   revokedAt: '2020-08-19T05:44:18.843Z',
   status: 'REVOKED'
 }
 
 export const completeConsentRevoked: Consent = {
-  id: '1234',
+  id: 'b51ec534-ee48-4575-b6a9-ead2955b8069',
   initiatorId: 'pisp-2342-2233',
   participantId: 'dfsp-3333-2123',
   status: 'REVOKED',
```
eoln, line 132: @lewisdaly if we drop `revoke` endpoint shouldn't we also update the `Consent.status` enumeration -> if we use DELETE method, maybe status should be DELETED not REVOKED. ?

eoln (changes requested):

src/interface/api-template.yaml
sridharvoruganti: Ok removed revoke and added new endpoint

src/interface/api-template.yaml
sridharvoruganti: removed generateChallenge endpoint

src/lib/db.ts
sridharvoruganti: these are all existing linting issues, will fix it in separate story.

src/lib/scopes.ts
sridharvoruganti: existing linting issues will be fixed in separate story.

src/server/handlers/consents.ts
sridharvoruganti: It's existing code not sure about that.

test/data/data.ts
sridharvoruganti: ok removed it

src/domain/consents.ts
sridharvoruganti: ok removed this interface, actually it's not required as  api-snippets used for types

test/data/data.ts
sridharvoruganti: removed it

src/server/create.ts
sridharvoruganti: existing linting issues will be fixed in separate story.

src/__mocks__/server.ts
```diff
@@ -26,7 +26,6 @@
  ******/
 
 // mock for server.run
-// eslint-disable-next-line @typescript-eslint/ban-types
-export default jest.fn((): object => ({
+export default jest.fn((): Record<string, unknown> => ({
```
eoln, line 29: simple workaround without introducing interface -> for mock it okay

src/server/handlers/consents.ts
```diff
@@ -49,11 +49,6 @@ export async function post (
...
   request: Request,
   h: ResponseToolkit): Promise<ResponseObject> {
-  // TODO: is this validation required? participantId removed from payload
-  // Validate request
-  // if (!isPostConsentRequestValid(request)) {
-  //   return h.response().code(Enum.Http.ReturnCodes.BADREQUEST.CODE)
-  // }
   // Asynchronously deals with creation and storing of consents and scope
   setImmediate(async (): Promise<void> => {
     try {
```
eoln, line 54: we need to drop using isMojaloopError here

src/server/handlers/index.ts
```diff
@@ -36,9 +34,7 @@ export default {
   HealthGet: Health.get,
   MetricsGet: Metrics.get,
   PostConsents: PostConsents.post,
-  RevokeConsent: RevokeConsent.post,
   UpdateConsent: UpdateConsent.put,
-  GenerateChallengeRequest: GenerateChallenge.post,
   VerifyThirdPartyAuthorization: Authorizations.post,
   validationFail: OpenapiBackend.validationFail,
   notFound: OpenapiBackend.notFound,
```
eoln, line 40: what about DELETE handler?

eoln: is it moved to another story?

eoln (changes requested):

src/server/handlers/consents.ts
sridharvoruganti: Ok dropped that comment code

src/server/handlers/index.ts
sridharvoruganti: Yes, as it's new endpoint will implement in new story also need functionality details like database updates.

src/server/handlers/consents.ts
sridharvoruganti: ok fixed

eoln (approved):

migrations/20200624121732_consent.ts
lewisdaly: You make a new migration :)

lewisdaly (approved):

PR #77, opened 2021-07-06
chore: update packages and regen audit resolve

kleyow:


src/domain/consents/ID.ts
```diff
@@ -77,29 +77,31 @@ export async function updateConsentCredential (
...
       credentialType: 'FIDO',
       status: CredentialStatusEnum.VERIFIED,
-      challenge: {
-        payload: consent.credentialChallenge as string,
-        signature: signature
-      },
-      payload: publicKey
+      payload: {
+        id: consent.credentialId!,
+        rawId: 'TODO: figure out what goes here',
```
eoln, line 98: `rawId` is a byte array made from string  `id`

src/model/scope/scope.ts
```diff
@@ -40,6 +40,8 @@ import { NotFoundError } from '../errors'
 
 /*
 * Interface for Scope resource type
+* TODO: rename this so it doesn't conflict with the Thirdparty API's
+*       Scope interface. Maybe something like AuthDBScope or ScopeDBRowObject
```
eoln, line 44: `ModelScope` ?

eoln (approved):

src/domain/consents/ID.ts
kleyow: Ah. Thank you. I'll get that in my pass that I'm working on today.

src/model/scope/scope.ts
kleyow: That works.

PR #78, opened 2021-07-08
chore: pull in redis/pub-sub/kvs/ttk/state plugin/state machine code for upcoming flows

kleyow:


docker-compose.yml
```diff
@@ -18,6 +18,9 @@ services:
       - mojaloop-net
     depends_on:
       - mysql
+      - redis
+    extra_hosts:
```
lewisdaly, line 22: Why do we need `extra_hosts`?

docker-compose.yml
```diff
@@ -46,4 +49,51 @@ services:
...
+      - "./docker/ml-testing-toolkit/secrets:/opt/mojaloop-testing-toolkit/secrets"
+    ports:
+      - "15000:5000"
+      - "5050:5050"
+    command: npm start
+    networks:
+      - mojaloop-net
+    depends_on:
+      - mongo
+    extra_hosts:
```
lewisdaly, line 77: Same as above. We don't need `extra _hosts`, it just understanding the networking more difficult.

lewisdaly (changes requested):
Looks good, just a minor comment on `extra_hosts`

docker-compose.yml
kleyow: Don't believe we need it. Just mindless copying. 
Let me know if I set up the compose files properly like the pisp repo.

lewisdaly: Cool. Yeah it looks fine to me apart from that section. We'll know if the config is wrong because the tests will break.

lewisdaly (approved):

PR #79, opened 2021-07-09
feat: implement register consent flow

kleyow:


src/model/registerConsent.model.ts
```diff
@@ -0,0 +1,328 @@
...
+            headers: {
+              Accept: 'application/vnd.interoperability.participants+json;version=1.1',
+              'FSPIOP-Source': this.config.authServiceParticipantFSPId,
+              Date: (new Date()).toUTCString()
+            }
+          }
+          const payload: fspiopAPI.Schemas.ParticipantsTypeIDSubIDPostRequest = {
+            fspId: this.config.authServiceParticipantFSPId
+          }
+          const res = await axios.post(alsParticipantURI, payload, axiosConfig)
```
lewisdaly, line 176: Why not use this? https://github.com/mojaloop/sdk-standard-components/blob/master/src/lib/requests/baseRequests.js#L307

test/unit/server/handlers/participants/{Type}/{ID}.test.ts
```diff
@@ -24,10 +24,12 @@
  - Kevin Leyow <kevin.leyow@modusbox.com>
  --------------
  ******/
-import { Request, ResponseToolkit } from '@hapi/hapi'
+import { Request } from '@hapi/hapi'
 import { Enum } from '@mojaloop/central-services-shared'
-import { h } from 'test/data/data'
 import ParticipantsTypeIDHandler from '~/server/handlers/participants/{Type}/{ID}'
+import { StateResponseToolkit } from '../../../../../../src/server/plugins/state';
```
lewisdaly, line 30: Any reason to use these instead of the `~`?

lewisdaly (approved):
Looks good, minor nitpicks.

I feel like all this time working on getting the state machines right is paying off.

test/unit/server/handlers/participants/{Type}/{ID}.test.ts
kleyow: auto import in vscode imports them like this and I sometimes miss it. 
been meaning to find a way to fix that.

src/model/registerConsent.model.ts
kleyow: We don't have type definitions for `BaseRequests` and instead of defining all the types this was easier until I can commit a  `postParticipantsTypeId` function to `sdk-standard-components`.

kleyow:
Going to merge this for now o/

PR #80, opened 2021-07-13
chore: add fido2-lib

kleyow:


lewisdaly:
@kleyow  I managed to get it working! There was some strange encoding stuff going on between `str2ab`, and what the front end client was doing.

It's now working, we just needed to base64encode the challenge in the `attestationExpectations` object in the end. I also fixed the front end client to encode the challenge to a `utf-8` ArrayBuffer (per the webauthn.guide example), instead of `utf-16`

kleyow:
Awesome. This saves us a ton of work. I can now write the registration/database logic/step into the RegisterConsent flow.

elnyry-sam-k (approved):

tsconfig.json
```diff
@@ -38,8 +38,8 @@
     "strictPropertyInitialization": true,
     "noImplicitThis": true,
     "alwaysStrict": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
```
lewisdaly, line 41: Can we put this back please? I was being lazy

lewisdaly (changes requested):

tsconfig.json
kleyow: done

lewisdaly (approved):

PR #85, opened 2021-08-13
refactor: restructure consent model for updated api

kleyow:


src/model/consent/consent.ts
```diff
@@ -42,45 +42,62 @@
...
 
-import { NotFoundError, RevokedConsentModificationError } from '../errors'
-import Knex from 'knex'
+import { NotFoundError } from '../errors'
+import Knex from 'knex';
 
 /*
  * Interface for Consent resource type
  */
 export interface Consent {
```
kleyow, line 51: So the `Consent` model was built around the idea of asking the auth-service for a challenge, storing a partial active consent and then later it would have a credential later be verified and updated on the `Consent` object.

Our API has changed to where the DFSP derives the challenge, PISP user signs and then the DFSP sends the `Consent` creation request alongside a credential to be verified.

So my thinking is 
- that the `credential*` fields should no longer be nullable since they are available on creation of the `Consent` object
- all credentials in `Consent` objects are verified since the `registerConsent` flow verifies before storing, bringing into question if `credentialStatus` is still needed
- `Consent` objects never need to be updated in the traditional sense. The only "updating" that happens is the revoking of the `Consent`. So I've removed all update logic that fell into the old logic of "create `Consent` object now, verify/update credential later"

The DB `schema` ditates a lot and I could use a second opinion @lewisdaly , as well as on any `TODO` you see as well.

Other than those refactoring. I've added two fields that we should need later in for assertion and remove fields that I don't think necessary anymore.

src/model/consent/consent.ts
```diff
@@ -42,45 +42,62 @@
...
+  participantId: string;
+  // status of the consent
   status: string;
-  credentialId?: string;
-  credentialType?: string;
-  credentialStatus?: string;
-  credentialPayload?: string;
-  credentialChallenge?: string;
+  // NOTE: not sure what purpose credentialId serves.
+  credentialId: string;
```
lewisdaly, line 58: I think this was the `keyHandleId` of the credential that is used to lookup the credential on the user's device.

src/model/consent/consent.ts
```diff
@@ -42,45 +42,62 @@
...
+  // credential type - currently trying to support FIDO/Generic credentials
+  credentialType: 'FIDO' | 'GENERIC' | undefined;
+  // NOTE: unsure this field is needed since `auth-service` verifies
+  //       credential on receiving them
+  credentialStatus: string;
+  // assuming this is the public key of the pair
+  credentialPayload: string;
```
lewisdaly, line 65: I think this should be type `string` | `FIDOPublicKeyCredentialAttestation`, depending on the `credentialType`

src/model/consent/consent.ts
```diff
@@ -42,45 +42,62 @@
...
-  credentialStatus?: string;
-  credentialPayload?: string;
-  credentialChallenge?: string;
+  // NOTE: not sure what purpose credentialId serves.
+  credentialId: string;
+  // credential type - currently trying to support FIDO/Generic credentials
+  credentialType: 'FIDO' | 'GENERIC' | undefined;
+  // NOTE: unsure this field is needed since `auth-service` verifies
+  //       credential on receiving them
+  credentialStatus: string;
```
lewisdaly, line 63: We should hold onto this, at least for auditing purposes.

In the API it can be `PENDING`, `VERIFIED` or `REVOKED`, but yes, in the database it should only really be `VERIFIED` or `REVOKED`

src/model/consent/consent.ts
```diff
@@ -42,45 +42,62 @@
...
   status: string;
-  credentialId?: string;
-  credentialType?: string;
-  credentialStatus?: string;
-  credentialPayload?: string;
-  credentialChallenge?: string;
+  // NOTE: not sure what purpose credentialId serves.
+  credentialId: string;
+  // credential type - currently trying to support FIDO/Generic credentials
+  credentialType: 'FIDO' | 'GENERIC' | undefined;
```
lewisdaly, line 60: remove `undefined` here

src/model/consent/consent.ts
```diff
@@ -42,45 +42,62 @@
...
+  credentialStatus: string;
+  // assuming this is the public key of the pair
+  credentialPayload: string;
+  // NOTE: not sure what `credentialChallenge` is used for.
+  //       best guess is that in `retrieveValidConsent` the original challenge
+  //       used to register the consent is needed to retrieve the Consent.
+  //       assuming this is an extra layer of protection since only the
+  //       DFSP has the variables used to derive the original challenge?
+  // TODO: remove this if it serves a redundant purpose
+  credentialChallenge: string;
```
lewisdaly, line 72: This is the original challenge sent by the DFSP to the PISP that is derived from the `scopes` + `consentId`: https://github.com/mojaloop/pisp-project/tree/master/docs/linking#161-deriving-the-challenge 

I think we should keep this as proof that the credential is indeed valid based on the `credentialPayload`

src/model/consent/consent.ts
lewisdaly: We can remove it, as it would be covered in `credentialPayload.id`

src/model/consent/consent.ts
lewisdaly: Ah I see. It's the consent.status that can be revoked, not the consent.credential status (ie. the entire consent is what is revoked, not just the credential of that consent)

So in that case, I think we can remove this field.

src/model/consent/consent.ts
```diff
@@ -42,45 +42,62 @@
...
  * Interface for Consent resource type
  */
 export interface Consent {
   id: string;
-  initiatorId?: string;
-  participantId?: string;
+  // participant DFSP that requested a Consent resource be made
+  participantId: string;
+  // status of the consent
   status: string;
```
lewisdaly, line 56: Should be of type `ConsentStatus`

lewisdaly (commented):
Just commenting on the Consent interface for now, once we are in agreement we can look more closely at the db

I'm also thinking we just scrap the old model and rewrite it... I'd rather that be much more readable than break the database for our 0 users.

src/model/consent/consent.ts
kleyow: `Consent` interface is 1:1 with the db are you suggesting we store all the fields from `FIDOPublicKeyCredentialAttestation`?

Looking at https://www.npmjs.com/package/fido2-lib I'm just assuming we need to store the public key from `await f2l.attestationResult` for FIDO at least

kleyow: I saw some old code that pointed me to the assumption that `credentialPayload` is a public key. Can't find it atm...

src/model/consent/consent.ts
kleyow: :+1:

src/model/consent/consent.ts
kleyow: Hmm `auth-service` used ACTIVE/REVOKED but since we are re-doing this might as well make it the same as the api currently i.e VERIFIED/REVOKED

src/model/consent/consent.ts
kleyow: sounds good

kleyow:
I believe you are commenting under the assumption that the `Consent` interface is akin to the tpAPI, but it really is a representation of the db model schema and is 1:1 atm. Maybe I'm wrong.

kleyow:
Think we might need to talk this over voice chat. @lewisdaly 
For now please take another look.

src/model/consent/consent.ts
lewisdaly: Yeah, you're right. The `f2l.attestationResult` should do the trick

src/model/consent/consent.ts
lewisdaly: Yes please.

lewisdaly:
Yeah I'm pretty lost here. 

Let's talk it over tonight before we dive into the thirdparty-scheme-adapter stuff

lewisdaly (approved):

PR #88, opened 2021-09-15
feat(verifyTransaction): implement fetch consent and verify transaction logic 

lewisdaly:
- clean up models and datatypes
- remove integration tests masquerading as unit tests

This PR is getting pretty big, so I'll fix and introduce the new Integration tests in the next one.


kleyow (approved):
I think you might have trouble with integration tests with sqllite and the `createdAt`/`revokedAt` fields. I can't quite remember why I decided to use `momentjs` but just keep that in mind if those fields start erroring.

lewisdaly:
sqlite is a bug... personally I think we should just remove it since nobody is actually going to use sqlite in production

PR #132, opened 2023-02-24
Feat/#3139 env set

kleyow:


.circleci/config.yml
```diff
@@ -264,7 +264,7 @@ jobs:
           command: mkdir -p ./audit/results
       - run:
           name: Check for new npm vulnerabilities
-          command: npm run audit:check --silent -- --json > ./audit/results/auditResults.json
+          command: npm run audit:check -- -o json > ./audit/results/auditResults.json
```
mdebarros, line 267: What happened here?

src/shared/config-db.ts
```diff
@@ -183,60 +85,153 @@ export const DatabaseConfigScheme = {
...
+    },
+    host: {
+      doc: 'Database host',
+      format: '*',
+      default: null,
+      env: 'DB_CONNECTION_HOST'
+    },
+    port: {
+      doc: 'Database port',
+      format: '*',
```
mdebarros, line 118:
```suggestion
      format: Number,
```

mdebarros (changes requested):
Snapshot is working fine for me.

Only suggested change is on the "format" for the port, which should be a number?

src/shared/config-db.ts
```diff
@@ -183,60 +85,153 @@ export const DatabaseConfigScheme = {
...
-    default: null
+    filename: {
+      doc: 'Database filename',
+      format: '*',
+      default: null,
+      env: 'DB_CONNECTION_FILENAME'
+    },
+    host: {
+      doc: 'Database host',
+      format: '*',
```
mdebarros, line 112:
```suggestion
      format: String,
```

src/shared/config-db.ts
```diff
@@ -183,60 +85,153 @@ export const DatabaseConfigScheme = {
...
+    },
+    port: {
+      doc: 'Database port',
+      format: '*',
+      default: null,
+      env: 'DB_CONNECTION_PORT'
+    },
+    user: {
+      doc: 'Database user',
+      format: '*',
```
mdebarros, line 124:
```suggestion
      format: String,
```

src/shared/config-db.ts
```diff
@@ -183,60 +85,153 @@ export const DatabaseConfigScheme = {
...
+      default: null,
+      env: 'DB_CONNECTION_USER'
+    },
+    password: {
+      doc: 'Database password',
+      format: '*',
```
mdebarros, line 130:
```suggestion
      format: String,
```

src/shared/config-db.ts
```diff
@@ -183,60 +85,153 @@ export const DatabaseConfigScheme = {
...
+      default: null,
+      env: 'DB_CONNECTION_PASSWORD'
+    },
+    database: {
+      doc: 'Database name',
+      format: '*',
```
mdebarros, line 136:
```suggestion
      format: String,
```

src/shared/config-db.ts
```diff
@@ -183,60 +85,153 @@ export const DatabaseConfigScheme = {
...
+      default: null,
+      env: 'DB_CONNECTION_NAME'
+    },
+    timezone: {
+      doc: 'Database timezone',
+      format: '*',
```
mdebarros, line 142: Do they have a format for timezone? or perhaps this should be an enum?

mdebarros (commented):
Some additional comments for your attention.

src/shared/config-db.ts
kleyow: No. Since `knex` services connecting to multiplate databases I think the format can be a varied set of strings.
Updated it to be a `String`

src/shared/config-db.ts
kleyow: done. here and throughtout.

kleyow:
@mdebarros addressed comments.

src/shared/config-db.ts
```diff
@@ -183,60 +85,160 @@ export const DatabaseConfigScheme = {
...
+      env: 'DB_CONNECTION_TIMEZONE',
+      nullable: true
+    }
   },
   pool: {
-    doc: 'Pool object specifying tarn pool properties',
-    format: DbPoolFormat.name,
-    default: null
+    min: {
+      doc: 'Database timezone',
```
mdebarros, line 156: Doc discriptions look like copy-pasta?

I think they all need to be updated to match the config pool paramater's description.

mdebarros (approved):
+1

